PR #66, opened 2022-07-22
Implement AWS SigV4 authentication

### LINKED ISSUE #44: IAM credentials to make connection to OS
Please provide a sample code that shows how to use IAM credentials to make connection to OS using from .NET client.

### LINKED ISSUE #65: [FEATURE] Add support for AWS SignV4 Request Signer for .NET
It would be nice to have the support for AWS SignV4 Request Signer for .NET
https://github.com/opensearch-project/opensearch-clients/issues/22

tfarrep:
### Description
Implements AWS SigV4 request signing, to enable authenticating with OpenSearch via AWS credentials.

Would be good to get feedback if there's anything to improve or better ways you'd like this integrated.

### Issues Resolved
Resolves #44 & Resolves #65 


MaxKsyunz:
Thank you @tfarrep!

dblock:
Without quite reading the code (hopefully @MaxKsyunz can CR this?), are we dragging a bunch of AWS SDK dependencies into .NET client at all times now? In the java client we've been trying to keep things optional/modular to avoid forcing client users to do that. Is it something possible with the .NET client?

tfarrep:
> Without quite reading the code (hopefully @MaxKsyunz can CR this?), are we dragging a bunch of AWS SDK dependencies into .NET client at all times now? In the java client we've been trying to keep things optional/modular to avoid forcing client users to do that. Is it something possible with the .NET client?

The AWS side of things is encapsulated into it's own separate library/package, so the client itself does not depend on it.

acarbonetto:
Related for IT tests: https://github.com/opensearch-project/opensearch-clients/issues/31

Yury-Fridlyand (changes requested):
Hi @tfarrep,

I checked your PR and I confirm that code works well. Unfortunately, I can't run unit or integration tests due to the test framework limitations, but manual test pass.
I have few notes to update there:
1. Add instructions and usage code samples. A simple readme.md file would be good enough.
2. Update tests to use .Net client test framework -- use `[U]` attribute instead of `[Fact]`.
3. You can also add new unit tests now. Unit tests run a virtual in-memory server. I works like a mirror -- replies `200 OK` or whatever you send to it. Have a look on `OpenSearch.Net.VirtualizedCluster` and `InMemoryConnection`.
4. According to @dblock's note it is proposed not to use `AWSSDK` libraries, but create a signature implementation which calls chained `HMAC` algorithm. Have a look on details [here](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html), [here](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-auth-using-authorization-header.html) and [here](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPOST.html).

Thank you!

tfarrep:
@Yury-Fridlyand Thanks for the feedback.

> 1. Add instructions and usage code samples. A simple readme.md file would be good enough.

Sure, what's your preferred location for that documentation/readme?

> 2. Update tests to use .Net client test framework -- use `[U]` attribute instead of `[Fact]`.

My bad, I'll fix those up.

> 3. You can also add new unit tests now. Unit tests run a virtual in-memory server. I works like a mirror -- replies `200 OK` or whatever you send to it. Have a look on `OpenSearch.Net.VirtualizedCluster` and `InMemoryConnection`.

If you could give me some more details on the type or level of testing desired, that'd be great. As there's not much validation I can see being achievable in memory, other validating the signing against itself (ie. signing the "received" request and ensuring it matches), or validating a specific request against a known signature value.

> 4. According to @dblock's note it is proposed not to use AWSSDK libraries, but create a signature implementation which calls chained HMAC algorithm. Have a look on details [here](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html), [here](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-auth-using-authorization-header.html) and [here](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPOST.html).

Fully removing any dependency on AWSSDK.Core, would make using the package quite un-wieldy, as you'd need to pass in an interface or Func for `GetCredentials`, which maps from AWSDK types to an OpenSearch owned type, in every project that use it. That credentials aren't "frozen" and construct time that we can just pass in simple strings.
I understood dblock's concern to be primarily about if it was pulling in the AWSDK dependency in the core client package.

dblock:
I am not an expert on this library, I'll let you guys figure out the best option here. I only came to comment here because I was working helping out with the Java version for Sigv4 and am interested in getting all OpenSearch clients to support it. More importantly, I'd like us to make sure we don't require users to have anything vendor-specific. The client should work equally well against an instance of OpenSearch on Azure or GCP without dragging in the AWS SDK kitchen sink.   

So I think from the developer POV I'd like to use this library as follows.

1) A project that knows nothing about AWS/SDK.

```
<PackageReference Include="OpenSearch.Net" Version="..." />
```

2) A project that wants to talk to Amazon OpenSearch and do Sigv4.

```
<PackageReference Include="OpenSearch.Net" Version="..." />
<PackageReference Include="OpenSearch.Net.Auth.AwsSigV4" Version="..." />
```

WDYT? What's possible?

tfarrep:
@dblock That's how I've implemented it at present. In this PR I've added a separate `OpenSearch.Net.AwsAuth` package (obviously can rename to be `.Auth.AwsSigV4` etc), which is where the dependency on `AWSSDK.Core` is fenced into. `OpenSearch.Net` or `OpenSearch.Client` etc know nothing of AwsSigV4.

MaxKsyunz:
@tfarrep could you amend [a8d17ab](https://github.com/opensearch-project/opensearch-net/pull/66/commits/a8d17abf4b90bf5e24684fd037666089211270c0) to include DCO sign-off as well?

MaxKsyunz:
@Yury-Fridlyand could you kick-off the workflows for this PR?

MaxKsyunz:
> @Yury-Fridlyand Thanks for the feedback.
> 
> > 1. Add instructions and usage code samples. A simple readme.md file would be good enough.
> 
> Sure, what's your preferred location for that documentation/readme?

@tfarrep  could you add a section to the [user guide](https://github.com/opensearch-project/opensearch-net/blob/main/USER_GUIDE.md)? There's already a section on how to use the library.

src/OpenSearch.Net.Auth.AwsSigV4/AwsSigV4HttpConnection.cs
```diff
@@ -0,0 +1,65 @@
+/* SPDX-License-Identifier: Apache-2.0
+*
+* The OpenSearch Contributors require contributions made to
+* this file be licensed under the Apache-2.0 license or a
+* compatible open source license.
+*
+* Modifications Copyright OpenSearch Contributors. See
+* GitHub history for details.
+*
+*  Licensed to Elasticsearch B.V. under one or more contributor
```
dblock, line 10: For new files you do not need to retain anything other than the SPDX header. We don't need to license this code explicitly to anyone since it's APLv2. See https://github.com/opensearch-project/.github/blob/main/CONTRIBUTING.md#license-headers.

dblock (commented):
I did a cursory look at the code, LGTM! You should cleanup license headers.

src/OpenSearch.Net.Auth.AwsSigV4/AwsSigV4HttpConnection.cs
tfarrep: The reason for them being that exact header, is there's a [workflow](https://github.com/opensearch-project/opensearch-net/blob/main/.github/workflows/license.yml) that runs this [script](https://github.com/opensearch-project/opensearch-net/blob/main/.github/check-license-headers.sh) to assert all `.cs` files have this [header](https://github.com/opensearch-project/opensearch-net/blob/main/.github/license-header.txt).

Additionally it's configured to be auto-added by IDEs [here](https://github.com/opensearch-project/opensearch-net/blob/main/OpenSearch.sln.DotSettings#L17-L43).

I'm more than happy to fix the license headers, but we need a decision on what to do with the workflow, I assume just expand it to check for either the "legacy" license or just the SPDX header?

dblock: I see. You can do this separately/before/after this PR - thank you. The workflow should check only SPDX, and we need to pay attention to not removing any headers from old(er) files in PRs.

tfarrep: I created a separate issue to discuss the desired approach here: https://github.com/opensearch-project/opensearch-net/issues/69

Yury-Fridlyand:
You can eliminate CI failures by merging main into your branch. Integration test workflow was updated recently.

USER_GUIDE.md
```diff
@@ -132,6 +132,54 @@ var response = lowLevelClient.Search<SearchResponse<Tweet>>("mytweetindex", Post
...
+### Configuring Region & Credentials
+By default, `AwsSigV4HttpConnection` will use the same default logic as the [AWS SDK for .NET](https://docs.aws.amazon.com/sdk-for-net/v3/developer-guide/creds-assign.html) to determine the credentials and region to use.
+However, you may explicitly specify one or both to override this logic, for example if you need to assume another IAM role first:
+```csharp
+var credentials = new AssumeRoleAWSCredentials(
+				FallbackCredentialsFactory.GetCredentials(),
+				"arn:aws:iam::123456789012:role/my-open-search-ingest-role",
+				"my-ingest-application");
+var region = RegionEndpoint.APSoutheast2;
+var connection = new AwsSigV4HttpConnection(credentials, region);
```
dblock, line 180: For completeness add the lines that create a client.

Next, many users have an endpoint and a region, but set credentials via AWS_ env variables. Document how to pass a region without credentials here and check whether this works:


```
set AWS_ACCESS_KEY_ID=
set AWS_SECRET_ACCESS_KEY=
set AWS_SESSION_TOKEN=
```

```
connect = new ... (host, region)
```

Maybe useful as another example?

src/OpenSearch.Net.Auth.AwsSigV4/AwsSigV4HttpClientHandler.cs
```diff
@@ -0,0 +1,42 @@
...
+	using System.Threading.Tasks;
+	using Amazon;
+	using Amazon.Runtime;
+
+	internal class AwsSigV4HttpClientHandler : DelegatingHandler
+	{
+		private readonly AWSCredentials _credentials;
+		private readonly RegionEndpoint _region;
+
+		public AwsSigV4HttpClientHandler(AWSCredentials credentials, RegionEndpoint region, HttpMessageHandler innerHandler)
```
dblock, line 24: I'd like default credentials but a configurable region.

src/OpenSearch.Net.Auth.AwsSigV4/AwsSigV4HttpConnection.cs
```diff
@@ -0,0 +1,45 @@
...
+	using System;
+	using Amazon;
+	using Amazon.Runtime;
+
+	public class AwsSigV4HttpConnection : HttpConnection
+	{
+		private readonly AWSCredentials _credentials;
+		private readonly RegionEndpoint _region;
+
+		public AwsSigV4HttpConnection(AWSCredentials credentials = null, RegionEndpoint region = null)
```
dblock, line 19: Are credentials scoped to a region, and thus do we need those separately here?

dblock (commented):
Nice work! Looks great, left some questions below. 

Before/after this is merged, typical questions will be whether this works with gzip compressed bodies, and chunked transfer-encoding. Will need to test/open issues.

USER_GUIDE.md
tfarrep: I've added further documentation in the direction of your comment here

src/OpenSearch.Net.Auth.AwsSigV4/AwsSigV4HttpClientHandler.cs
tfarrep: This particular class is a fully internal implementation detail where the credentials and region are already known.

src/OpenSearch.Net.Auth.AwsSigV4/AwsSigV4HttpConnection.cs
tfarrep: I'm not sure I understand your question or intent here? No, credentials aren't tied to a region, it'd be perfectly possible to do something like this contrived example:
```csharp
var credentials = new AssumeRoleAWSCredentials(
				FallbackCredentialsFactory.GetCredentials(),
				"arn:aws:iam::123456789012:role/my-open-search-ingest-role",
				"my-ingest-application");

var usClient = new OpenSearchClient(
                new ConnectionSettings(
                    new Uri("https://example-aaabbbcccddd111222333.us-east-1.es.amazonaws.com"), 
                    new AwsSigV4HttpConnection(credentials, RegionEndpoint.USEast1)));

var apClient = new OpenSearchClient(
                new ConnectionSettings(
                    new Uri("https://example-aaabbbcccddd111222333.ap-southeast-2.es.amazonaws.com"), 
                    new AwsSigV4HttpConnection(credentials, RegionEndpoint.APSoutheast2)));
```

tfarrep:
> Nice work! Looks great, left some questions below.
> 
> Before/after this is merged, typical questions will be whether this works with gzip compressed bodies, and chunked transfer-encoding. Will need to test/open issues.

I'll give gzip compressed bodies and chunked transfer encoding a look. The current implementation doesn't account for chunked transfer, so my first guess is that it won't _break_ it'll just not do chunked encoding.

src/OpenSearch.Net.Auth.AwsSigV4/AwsSigV4HttpConnection.cs
dblock: Can I just use `FallbackCredentialsFactory.GetCredentials()` as `credentials`? I *think* you need to get something with a region here, but I am not sure.

tfarrep: Ah, with how the constructor was here it'd be `new AwsSigV4HttpConnection(region=RegionEndpoint.USEast1)`. But I've now made the additional overloads explicit for the constructor, so your example would work now.

lizsnyder:
Will we be able to upload this to NuGet once it's merged?

Yury-Fridlyand:
> Will we be able to upload this to NuGet once it's merged?

For sure, this will be a part of a release.

dblock:
This LGTM! Rebase/fix license headers check? @Yury-Fridlyand all yours to merge!

dblock:
Let's get this back to rebased 🟢  and merge?

Xtansia:
> Let's get this back to rebased 🟢 and merge?

@dblock This needs #78 to be merged first before it'll pass all checks

Yury-Fridlyand:
Can you update license headers to comply with the new format?
https://github.com/opensearch-project/opensearch-net/runs/7846155570?check_suite_focus=true

dblock:
> > Let's get this back to rebased 🟢 and merge?
> 
> @dblock This needs #78 to be merged first before it'll pass all checks

Looks like that made it in now, thanks!

Xtansia:
@dblock @Yury-Fridlyand It has now been rebased.

Yury-Fridlyand:
> @dblock @Yury-Fridlyand It has now been rebased.

Compilation fails. You have to update `packages.lock.json` files.
You can delete these files at our local clone, the run `dotnet build` and these files would be regenerated.
Please, ensure that they don't have unnecessary changes and have `eol = lf` (not `crlf`).

tests/Tests.Auth.AwsSigV4/CanonicalRequestTests.cs
```diff
@@ -0,0 +1,123 @@
...
+{
+	using System;
+	using System.Linq;
+	using System.Net.Http;
+	using System.Threading.Tasks;
+	using System.Web;
+	using Amazon.Runtime;
+	using FluentAssertions;
+	using OpenSearch.Net.Auth.AwsSigV4;
+	using OpenSearch.OpenSearch.Xunit.XunitPlumbing;
```
Yury-Fridlyand, line 18: Why `using` block is inside `namespace`?

Yury-Fridlyand:
Thank you for fixing unit tests!

tests/Tests.Auth.AwsSigV4/CanonicalRequestTests.cs
Xtansia: Will just be what my IDE was defaulted to due to personal preference, hadn't realised there wasn't a specific setting for this project and that it'd ended up mismatched with the rest of the codebase. Happy to move them outside the namespace if that's desired.

Yury-Fridlyand: I think it is good to move it outside for consistency with the rest project. Thanks!

dblock (dismissed):
This LGTM, but leaving it to @Yury-Fridlyand

Yury-Fridlyand:
Fine for me too, but let us finish with the upcoming release first.

MaxKsyunz:
@tfarrep @Yury-Fridlyand feature branch now conflicts with main.

Xtansia:
@Yury-Fridlyand @dblock Do we have any blockers on getting this re-reviewed (and hopefully merged)?

Yury-Fridlyand (approved):

Yury-Fridlyand:
No! 🚀

dblock (approved):

dcsena:
Thanks for getting this done! Can someone share deployment schedule? When can we expect to consume this via nuget?

wbeckler:
Since a release just happened, this might not go into nuget for a while, but feel free to download it and run it anyway. I'm pretty sure nobody will break this in the next release.

PR #93, opened 2022-09-06
Bump the version.

Yury-Fridlyand:
Signed-off-by: Yury-Fridlyand <yuryf@bitquilltech.com>

### Description
Bump release version to `1.1.0`.


dblock (approved):
We should be doing this every time _after_ a release, preparing for the next developer iteration.

Yury-Fridlyand:
This goes to the `main`, not to `1.1.0` branch.

harshavamsi:
> We should be doing this every time _after_ a release, preparing for the next developer iteration.

@dblock correct me if I'm wrong here, but shouldn't this be done _before_ a release. The release artifacts still point to the older `1.0.0` version unless we update them here first.

wbeckler (approved):

dblock:
@harshavamsi My favorite reason why version should be incremented to the next development iteration right _after_ a release is because I don't ever want to rebuild code with changes that happened after the said release. So let's say we release 1.0.0, the next patch version is always 1.0.1 per semver (incremented immediately) and there's never again a 1.0.0 build.

Issue #125, opened 2022-12-12
[PROPOSAL] Retire the net461 target & migrate to .NET 6 SDK


Xtansia:
### What are you proposing?
I propose removing the explicit `net461` target from our packages as [.NET Framework 4.6.1 reached End of Support in April 2022](https://learn.microsoft.com/en-gb/lifecycle/faq/dotnet-framework), and additionally updating the SDK used to build the packages to .NET 6 as [.NET 5 reached End of Support in May 2022](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core).

### What problems are you trying to solve?
- Due to .NET Framework 4.6.1 reaching EoS, more and more packages (especially those published by Microsoft) are removing it from their supported targets, which results in issues for us when we want to update our own dependencies.
- Continuing to use .NET 5 as our build SDK may result in security issues due to being at EoS, and it limits support for our test packages and other tools. Additionally .NET 6 supports the Apple M1 processor, meaning a better developer experience for any who may be using an M1 based machine.

### What is the developer experience going to be?
- Increasing the required SDK version will mean developers will need to install the newer SDK if not already installed, otherwise the experience will be unchanged.

#### Are there any security considerations? 
- If anything this will improve the security situation by enabling us to stay better up to date

### What is the user experience going to be?
Consumers of the library will only be impacted if they are specifically targeting .NET Framework 4.6.1, all other consumers (including those who have updated to .NET Framework 4.6.**2** from 4.6.1) will be unaffected as [all other recent versions of .NET Framework and .NET support .NET Standard 2.0](https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0) which we target. 

#### Are there breaking changes to the User Experience?
As above regarding those on 4.6.1

### Why should it be built? Any reason not to?
Enables continued modernisation and package updates of the client, resulting in better user experiences and improved security posture.

### What will it take to execute?
Requires minimal effort, updating SDK version number in global.json and workflows, updating necessary packages, removing net461 target from projects.

### Any remaining open questions?
- Dropping the net461 target likely implies a major semver bump due to it being a breaking change for a (very small if any) percentage of users.

Xtansia:
Additionally this can be done as separate stages:
1. Update required .NET SDK for compiling to 6
2. Remove net461 target from projects only used internally within the repo that aren't published
3. Remove net461 target from the main client projects

PR #128, opened 2022-12-21
Remove net461 target from internal packages

Xtansia:
### Description
Remove the net461 target from internal packages and update a couple dependencies which have CVEs. The client itself does not depend on these and is unaffected by the CVEs.


dblock (approved):
Any downsides to this?

Xtansia:
> Any downsides to this?

@dblock As these are purely internal packages and unpublished, so this doesn't break any published API, and `net461` target on these packages was unused internally.

The main driving factor here is being able to keep dependencies up to date https://github.com/opensearch-project/opensearch-net/issues/125

dblock (approved):

Yury-Fridlyand (approved):

Issue #129, opened 2022-12-23
[FEATURE] Support for AWS OpenSearch Serverless Request Signing


zacharyhall:
### Is your feature request related to a problem?

I would like to leverage the ease of management with the AWS Open Search Serverless service for our project without writing my own client/request signing implementation; however, this client currently does not seem to support signing requests for this new service yet. 

### What solution would you like?

The main issue seems to lay [here](https://github.com/opensearch-project/opensearch-net/blob/da0f0ac8012ca8ec1a7a13deaf8eb982910fb9af/src/OpenSearch.Net.Auth.AwsSigV4/AwsSigV4Util.cs#L24), in this bit of code:
``` C#
var canonicalRequest = await CanonicalRequest.From(request, credentials, signingTime).ConfigureAwait(false);

var signature = AWS4Signer.ComputeSignature(credentials, region.SystemName, signingTime, "es", canonicalRequest.SignedHeaders,
canonicalRequest.ToString());

request.Headers.TryAddWithoutValidation("x-amz-date", canonicalRequest.XAmzDate);
request.Headers.TryAddWithoutValidation("authorization", signature.ForAuthorizationHeader);
if (!string.IsNullOrEmpty(canonicalRequest.XAmzSecurityToken)) request.Headers.TryAddWithoutValidation("x-amz-security-token", canonicalRequest.XAmzSecurityToken);
```

`SignRequests` and/or anything that invokes `SignRequests`, could potentially be overloaded to either be configurable or to explicitly handle requests and signing for OpenSearch Serverless. Otherwise, different handler/connection logic could potentially be created just for the Serverless service (e.g. OpenSearch.Net.Serverless.Auth.AwsSigV4)

The key changes required to allow for OpenSearch Serverless requests seem to be:

1. The Service Name ("es") in [this line](https://github.com/opensearch-project/opensearch-net/blob/da0f0ac8012ca8ec1a7a13deaf8eb982910fb9af/src/OpenSearch.Net.Auth.AwsSigV4/AwsSigV4Util.cs#L32) needs to be "aoss":
``` C#
var signature = AWS4Signer.ComputeSignature(credentials, region.SystemName, signingTime, "es", canonicalRequest.SignedHeaders, canonicalRequest.ToString());
```
[...]

PR #133, opened 2023-01-12
Support AwsSigV4 signing for Amazon OpenSearch Serverless

Xtansia:
### Description
Support AwsSigV4 signing for Amazon OpenSearch Serverless
- Make the service ID customizable, so it can be set to "aoss" for serverless
- Include the x-amz-content-sha256 header
- Don't sign the content-length header

Relates to #129, however does not solve documenting of methods unsupported by serverless.

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -10,6 +10,7 @@
     - [Getting Started](#getting-started-1)
     - [Connecting](#connecting-1)
     - [Configuring Region & Credentials](#configuring-region--credentials)
+    - [Amazon OpenSearch Serverless](#amazon-opensearch-serverless)
   - [OpenSearch.Net](#opensearchnet)
     - [Getting Started](#getting-started-2)
     - [Connecting](#connecting-2)
@@ -26,7 +27,7 @@ Include OpenSearch.Client in your .csproj file.
 <Project>
   ...
   <ItemGroup>
-    <PackageReference Include="OpenSearch.Client" Version="1.0.0" />
+    <PackageReference Include="OpenSearch.Client" Version="1.*" />
   </ItemGroup>
 </Project>
 ```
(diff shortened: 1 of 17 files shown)


USER_GUIDE.md
```diff
@@ -219,6 +220,21 @@ var config = new ConnectionSettings(endpoint, connection);
 var client = new OpenSearchClient(config);
 ```
 
+### Amazon OpenSearch Serverless
+To configure signing when making requests to [Amazon OpenSearch Serverless](https://aws.amazon.com/opensearch-service/features/serverless/) is nearly identical to all above configuration for AwsSigV4, the only difference being the need to configure the service identifier.
```
dblock, line 224: Nit: Maybe not so english :) How about "Use the `oass` service identifier to make requests to ..."?

dblock (approved):
This looks good.

Is there any different between managed service and serverless implementation anywhere and have you tested both with this change, including with POSTing a request body (add data)? There shouldn't be any variation between the two.

Xtansia:
> This looks good.
> 
> Is there any different between managed service and serverless implementation anywhere and have you tested both with this change, including with POSTing a request body (add data)? There shouldn't be any variation between the two.

Have done some testing using BulkIndex and Search on both managed service and serverless with these changes. There's no branching in logic between the two in signing, only difference being the service ID.

Yury-Fridlyand (approved):

Yury-Fridlyand:
Open test failed for `ArtifactsApi`. It is not related for the changes, but please consider fixing it (in another PR).

src/OpenSearch.Net.Auth.AwsSigV4/AwsSigV4HttpClientHandler.cs
```diff
@@ -20,19 +20,21 @@ internal class AwsSigV4HttpClientHandler : DelegatingHandler
 	{
 		private readonly AWSCredentials _credentials;
 		private readonly RegionEndpoint _region;
+		private readonly string _serviceId;
```
dblock, line 23: According to https://docs.aws.amazon.com/general/latest/gr/signing-elements.html this should be called "service code". I see other clients use "service" or "service name". At least I think we shouldn't introduce a new one 😅

Yury-Fridlyand (approved):

dblock (dismissed):

dblock:
This could use some tests that the service name is passed through.

dblock (commented):
Looking good if you can bring it to 🟢.

Yury-Fridlyand (approved):

dblock (approved):

Yury-Fridlyand (approved):

PR #216, opened 2023-05-17
Implement code generation from OpenAPI spec

### LINKED ISSUE #193: [FEATURE] Generate API code from OpenSearch API Spec
Related to https://github.com/opensearch-project/opensearch-clients/issues/19

Re-use the generator that generates API Code that maintains backwards compatibility with the current client APIs and provides a way to add new APIs.

Old generator was cleaned up in https://github.com/opensearch-project/opensearch-net/commit/7db02e979b19c748879b74485f04f6c352a0d967 so will need reverting and modification.

Using OpenAPI generators instead of re-using the existing generator but it makes it difficult to maintain backwards compatibility.

Xtansia:
### Description
Brings back the old `ApiGenerator` and refactors it to instead base off an OpenAPI spec

### Issues Resolved
Closes #193 


Yury-Fridlyand:
What was changed in API generator? Can I skip first commit for reviewing?

Xtansia:
> What was changed in API generator? Can I skip first commit for reviewing?

First commit is just a direct restore, so can be skipped. Main changes in ApiGenerator were general cleanup and refactoring plus the changes to the api spec parsing. There'll need to be a little more change to make it download the spec from the correct place

Xtansia:
Can see an example of re-generating the tasks namespace here: https://github.com/Xtansia/opensearch-net/compare/feat/code-generation...Xtansia:opensearch-net:generation/tasks

Xtansia:
I've decided to break this into smaller PRs with fewer changes to make it easier to review. Starting with just the revert commit in #222

PR #228, opened 2023-06-14
Implement parsing OpenAPI spec in ApiGenerator

Xtansia:
### Description
Implement parsing of the OpenAPI spec in ApiGenerator

### Issues Resolved
Implements part of #193 


dblock (approved):
I did a cursory review and LGTM.

src/ApiGenerator/OpenSearch.openapi.json
VachaShah, line 1: Can we use the file dynamically from the `opensearch-api-specification` repo? What happens when the OpenAPI spec changes?

Xtansia: This is the same behaviour as previously with legacy rest specs. it will/can download the spec from the `opensearch-api-specification` repo. It's just optional behind a CLI flag and keeps a copy of the most recently used spec so you can re-generate if you need to only change templates without pulling newer spec.

VachaShah (approved):

PR #256, opened 2023-07-05
Remove support for the `net461` target

### LINKED ISSUE #125: [PROPOSAL] Retire the net461 target & migrate to .NET 6 SDK

Xtansia:
### Description
Removes support for the `net461` target from the client. As explained in https://github.com/opensearch-project/opensearch-net/issues/125 `net461` is out-of-support, and 4.6.2 is supported via the `.netstandard2.0` target.

### Issues Resolved
Resolves #125 


dblock (approved):

dblock:
Do you want to backport this anywhere?

Xtansia:
@dblock no as this is a breaking change, it shouldn't go into 1.x

PR #265, opened 2023-07-07
Remove packages.lock.json & fix CI caching

Xtansia:
### Description
Remove packages.lock.json as they generally are more effort to maintain than they offer in benefits. Fix caching in CI builds as previously nuget cache was being immediately wiped after being restored.


.github/workflows/compile.yml
```diff
@@ -13,14 +13,6 @@ jobs:
         dotnet-version: |
           5.0.x
           6.0.x
-    - uses: actions/cache@v2
-      with:
-        path: ~/.nuget/packages
```
Yury-Fridlyand, line 18: You can cache/restore packages (without lock files) according to the version.
Without cache `dotnet` will download all dependencies in every CI job.

Yury-Fridlyand (dismissed):

.github/workflows/integration.yml
```diff
@@ -37,23 +37,19 @@ jobs:
...
 
-      - uses: actions/setup-dotnet@v1
+      - uses: actions/setup-dotnet@v3
         with:
           dotnet-version: |
             5.0.x
             6.0.x
-
-      - uses: actions/cache@v2
+
```
Yury-Fridlyand, line 45: nit
```suggestion

```

.github/workflows/build_deploy_docs.yml
```diff
@@ -18,6 +18,13 @@ jobs:
             5.0.x
             6.0.x
 
+      - uses: actions/cache@v3
+        with:
+          path: ~/.nuget/packages
+          key: ${{ runner.os }}-nuget-${{ hashFiles('**/*.csproj') }}
```
Yury-Fridlyand, line 24: What about F# project files? Does this glob work?
```suggestion
          key: ${{ runner.os }}-nuget-${{ hashFiles('**/*.*sproj') }}
```

.github/workflows/build_deploy_docs.yml
```diff
@@ -18,6 +18,13 @@ jobs:
...
+          restore-keys: |
```
Yury-Fridlyand, line 25: Does it save cache ever? 🤔 Did is save it in past? 🤔 🤔 
My recent implementation of cache save/load: https://github.com/opensearch-project/sql-odbc/blob/2f43282431704e10b33de37a3d8046e2c98a339b/.github/workflows/sql-odbc-main.yml#L41-L61 (not a best example though)

Xtansia: You can see it save here https://github.com/opensearch-project/opensearch-net/actions/runs/5502227861/jobs/10026345514?pr=265#step:8:4 (and restoring from an earlier cache here https://github.com/Xtansia/opensearch-net/actions/runs/5502227521/jobs/10026345103#step:4:17)

The only difference between this and the explicit restore/save steps, is that it doesn't try to save until after the workflow has fully succeeded, it doesn't save early.

Yury-Fridlyand (approved):

PR #303, opened 2023-08-04
Run rest-api-spec / YAML tests in CI

Xtansia:
### Description
Fixes the test runner to download specs in a manner that won't get rate-limited. 
Add CI jobs to run the YAML tests


.github/actions/build-opensearch/action.yml
```diff
@@ -0,0 +1,103 @@
...
+      uses: ./client/.github/actions/cached-git-build
+      if: inputs.security_plugin == 'true'
+      with:
+        repository: opensearch-project/security
+        ref: ${{ inputs.ref }}
+        path: opensearch-security
+        cache_key_suffix: ${{ inputs.build_snapshot == 'true' && '-snapshot' || '' }}
+        cached_paths: |
+          ./opensearch-security/build/distributions/opensearch-security-*.zip
+        build_script: ./gradlew assemble -Dopensearch.version=${{ steps.determine.outputs.version }} -Dbuild.snapshot=${{ inputs.build_snapshot }}
```
Yury-Fridlyand, line 65: Why not to download snapshot from maven?
https://aws.oss.sonatype.org/content/repositories/snapshots/org/opensearch/plugin/opensearch-security/3.0.0.0-SNAPSHOT/

.github/workflows/integration.yml
```diff
@@ -115,85 +90,138 @@ jobs:
...
+          OPENSEARCH_DISTRIBUTION: ${{ steps.opensearch.outputs.distribution }}
         
       - name: Upload test report
         if: failure()
         uses: actions/upload-artifact@v3
         with:
           name: report-unreleased-${{ matrix.opensearch_ref }}
           path: client/build/output/*
+
+  test-yaml:
```
Yury-Fridlyand, line 117: Do you want to add to a new workflow?

.github/workflows/test-jobs.yml
```diff
@@ -39,20 +39,13 @@ jobs:
...
+        uses: mikepenz/action-junit-report@v3
         with:
           report_paths: 'build/output/junit-*.xml'
           github_token: ${{ secrets.GITHUB_TOKEN }}
           fail_on_failure: true
           require_tests: true
           check_name: Unit Test Results
-        # This is the only way to get test reports out from GitHub Actions
-        # TODO remove this before release
-      - uses: actions/upload-artifact@v2
```
Yury-Fridlyand, line 51: Still could be useful to investigate failures. I hope you never have one though.

Yury-Fridlyand (dismissed):

.github/actions/build-opensearch/action.yml
Xtansia: I'd missed one of the conditions, but the security plugin is only built if we're testing against `1.x`, as snapshots for that aren't published to maven, otherwise it is normally downloaded from maven.

k-NN needs to be built as the published snapshots do not contain the native libs, and also do not exist for `1.x`.

.github/workflows/test-jobs.yml
Xtansia: I've re-added these, just changed the condition to `failed()` rather than `always()`

VachaShah (approved):
Some tests are failing, are they flaky tests?

Xtansia:
> Some tests are failing, are they flaky tests?

The tests for unreleased look to be failing due to changes in core that haven't made it into k-NN. 
Test against 2.9.0 is getting a 500 error for one set of tests, need to check if it's a change in behaviour or bug in core.

Will create issues for them and investigate

PR #322, opened 2023-08-17
Fixes #321: fix highlight max_analyzer_offset field name to match OS 2.2.0

gjunge:
Fix field `max_analyzed_offset` to `max_analyzer_offset`  

### Description
The current field `max_analyzed_offset` used in ElasticSearch doesn't match the field added to OpenSearch version 2.2.0, `max_analyzer_offset`.
This means the current client implementation have two problems:

It doesn't work for versions before version 2.2.0, because no equivalent field exist on OpenSearch.
It doesn't work for version after 2.2.0, because the field is misspelled.

Even though the Property changes, which is normally a breaking change, since the current field is unusable, this change is not "breaking" anything.

For further information see:
- https://github.com/opensearch-project/opensearch-java/pull/555
- https://github.com/opensearch-project/OpenSearch/pull/4031
- https://github.com/opensearch-project/OpenSearch/pull/3893

### Issues Resolved:

- #321 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #324, opened 2023-08-20
[Backport 1.x] Fixes #321: fix highlight max_analyzer_offset field name to match OS …

### LINKED ISSUE #321: [BUG] "Misspelling" on setting MaxAnalyzedOffset on IHighlight
### What is the bug?
Same bug as https://github.com/opensearch-project/opensearch-java/issues/553 , since this part was copied from the ES clients. ES implemented this field as `max_analyzed_offset` ([see ES docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/highlighting.html)), while OpenSearch implemented the field as `max_analyzer_offset` in 2.2.0  ([see OS docs](https://opensearch.org/docs/2.2/opensearch/search/highlight/))

As can be seen in the Java client bug description, using this field will lead to this error

```
    "error": {
        "root_cause": [
            {
                "type": "x_content_parse_exception",
                "reason": "[36:9] [highlight] unknown field [max_analyzed_offset] did you mean [max_analyzer_offset]?"
            }
        ],
        "type": "x_content_parse_exception",
        "reason": "[36:9] [highlight] unknown field [max_analyzed_offset] did you mean [max_analyzer_offset]?"
    },
    "status": 400
}
```

```
var results = await _opensearchClient.SearchAsync<MyRecord>(
        sd => sd.TrackTotalHits(true)
                .Query(q => q.MatchAll())
                .Highlight(h =>
                h.PreTags("<mark>")
                    .PostTags("</mark>")
                    .Fields(
                        f => f.Field("title")
                    )
                    .MaxAnalyzedOffset(10000)  //<-- this is the issue.
                )


        );
```
Update the api to match the OpenSearch Highlight setting max_analyzer_offset

AWS Managed OpenSearch service 2.7.

Xtansia:
### Description
Backports 8ad8057696044485d53858b794f2d6d0dc74aa8d from #322 

### DIFF
--- a/src/OpenSearch.Client/Search/Search/Highlighting/Highlight.cs
+++ b/src/OpenSearch.Client/Search/Search/Highlighting/Highlight.cs
@@ -107,8 +107,9 @@ namespace OpenSearch.Client
 		/// If this setting is set to a non-negative value, the highlighting stops at this defined maximum limit, and the
 		/// rest of the text is not processed, thus not highlighted and no error is returned.
 		/// </summary>
-		[DataMember(Name ="max_analyzed_offset")]
-		int? MaxAnalyzedOffset { get; set; }
+		/// <remarks>Introduced in OpenSearch 2.2</remarks>
+		[DataMember(Name ="max_analyzer_offset")]
+		int? MaxAnalyzerOffset { get; set; }
 
 		[DataMember(Name ="max_fragment_length")]
 		int? MaxFragmentLength { get; set; }
@@ -210,7 +211,7 @@ namespace OpenSearch.Client
 		public QueryContainer HighlightQuery { get; set; }
 
 		/// <inheritdoc/>
-		public int? MaxAnalyzedOffset { get; set; }
+		public int? MaxAnalyzerOffset { get; set; }
 
 		/// <inheritdoc/>
 		public int? MaxFragmentLength { get; set; }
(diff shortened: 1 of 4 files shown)


src/OpenSearch.Client/Search/Search/Highlighting/Highlight.cs
```diff
@@ -107,8 +107,9 @@ public interface IHighlight
 		/// If this setting is set to a non-negative value, the highlighting stops at this defined maximum limit, and the
 		/// rest of the text is not processed, thus not highlighted and no error is returned.
 		/// </summary>
-		[DataMember(Name ="max_analyzed_offset")]
-		int? MaxAnalyzedOffset { get; set; }
+		/// <remarks>Introduced in OpenSearch 2.2</remarks>
+		[DataMember(Name ="max_analyzer_offset")]
```
Yury-Fridlyand, line 111: It should have both `max_analyzer_offset` and `max_analyzed_offset` for compatibility with pre 2.2 and post 2.2 versions.

Xtansia: This option did not exist in OpenSearch pre 2.2 with either spelling, so the option would not have functioned at all before. It only would have working if using the client against ES versions that have it with the `ed` spelling it but we make no guarantees of compatibility with ES. Which is why I'm inclined to say this is okay for a minor version bump rather than major. If we think it's important to keep the old spelling to not break those using the client against ES then I can re-add the old property as deprecated in 1.x for removal in 2.x/main.

@`2.1`: https://github.com/opensearch-project/OpenSearch/blob/2.1/server/src/main/java/org/opensearch/search/fetch/subphase/highlight/AbstractHighlighterBuilder.java#L73-L94
@`2.2`: https://github.com/opensearch-project/OpenSearch/blob/2.2/server/src/main/java/org/opensearch/search/fetch/subphase/highlight/AbstractHighlighterBuilder.java#L73-L96

Yury-Fridlyand: Thank you for clarifying things!

Yury-Fridlyand (approved):

PR #405, opened 2023-10-19
Add support for Point In Time APIs

### LINKED ISSUE #220: [FEATURE] Support for Point in time search
I am trying to find a way to use the OS .Net client to create and delete PITs and to do search against a PIT ID but the client v1.3 does not support that. Is there a plan to add support for Open Search v2.4 or higher so PITs can be supported? If not, is there a neat way to extend the client to achieve this?

I also thought about using a standard HttpClient and invoke the PIT APIs directly but there are obvious risks with that.

Xtansia:
### Description
Adds support for the Point In Time APIs.

### Issues Resolved
Resolves #220 


Xtansia:
@dblock @Yury-Fridlyand @VachaShah Could I get a review on this please?

Issue #403, opened 2023-10-16
[FEATURE] Make raw JSON REST requests to OpenSearch


dblock:
### Is your feature request related to a problem?

Coming from https://github.com/opensearch-project/opensearch-clients/issues/62, add support for making raw JSON requests.

### What solution would you like?

1. If an existing mechanism exists, document it similar to https://github.com/opensearch-project/opensearch-py/pull/542. 
2. Add a high level DSL, e.g. `client.get` and `client.post`.

Xtansia:
This exists in the low-level OpenSearch.Net client, but there's no high-level DSL wrapper. Example of its use can be seen here: https://github.com/opensearch-project/opensearch-net/issues/220#issuecomment-1571146526

dblock:
```cs
await client.LowLevel.DoRequestAsync<VoidResponse>(
            HttpMethod.DELETE,
            "_search/point_in_time",
            CancellationToken.None,
            PostData.Serializable(new
            {
                pit_id = pitId
            }));
```

Nice. Documenting it would be a great start!

Djcarrillo6:
Hi @Xtansia, I'd like to request assignment on this issue

dblock:
Does it make sense to alias `LowLevel` to `HTTP` or `http`?

https://github.com/opensearch-project/opensearch-net/pull/406#issuecomment-1808209065

Xtansia:
> Does it make sense to alias `LowLevel` to `HTTP` or `http`?
> 
> [#406 (comment)](https://github.com/opensearch-project/opensearch-net/pull/406#issuecomment-1808209065)

A direct alias of `LowLevel` doesn't make much sense as it points to an instance of the full low-level client, meaning you'd get `client.Http.Indices.PutTemplate`. 

I have some ideas for a high-level DSL for the http methods, that I'm going to look into feasibility for.

PR #406, opened 2023-10-22
Added guide for using raw JSON REST requests with low-level client.

Djcarrillo6:
### Description
Adds a new guide that demonstrates how to interact with the OpenSearch REST API using the OpenSearch .NET low-level client by executing raw JSON REST requests. It covers a number of operations such as setting up an OpenSearch instance, indexing document data, and executing CRUD operations (CREATE, READ, UPDATE and DELETE) on documents in the index.

### Issues Resolved
Resolves issue [403](https://github.com/opensearch-project/opensearch-net/issues/403)


guides/using-json-rest-requests.md
```diff
@@ -0,0 +1,135 @@
+# Using Raw JSON REST Requests
```
dblock, line 1: Let's try to match the format in https://github.com/opensearch-project/opensearch-py/blob/main/guides/json.md?

guides/using-json-rest-requests.md
```diff
@@ -0,0 +1,135 @@
...
+    public string Title { get; set; }
+    public string Director { get; set; }
+    public int Year { get; set; }
+    public override string ToString()
+    {
+        return $"{nameof(Title)}: {Title}, {nameof(Director)}: {Director}, {nameof(Year)}: {Year}";
+    }
+}
+
+private class CreatePitParams : RequestParameters<CreatePitParams>
```
dblock, line 29: This looks unused.

guides/using-json-rest-requests.md
```diff
@@ -0,0 +1,135 @@
...
+    public string KeepAlive
+    {
+        get => Q<string>("keep_alive");
+        set => Q("keep_alive", value);
+    }
+}
+
+var index = "movies";
+
+await client.IndexManyAsync(
```
dblock, line 43: For a pure JSON example you can't use `IndexManyAsync`.

dblock (commented):
Good start.

1. Remove any usage of higher level classes that aren't raw request/response.
2. Write a working sample and put it into `samples` so we can make sure this code works.

guides/using-json-rest-requests.md
Djcarrillo6: @dblock I don't think this repo has a sample directory like the Python client does. Should I create a sample dir in this .NET client repo & put it there? Or did you mean you want me to put the sample somewhere else as in another repo?

Djcarrillo6:
> Good start.
> 
> 1. Remove any usage of higher level classes that aren't raw request/response.
> 2. Write a working sample and put it into `samples` so we can make sure this code works.

@dblock @Xtansia If I'm not mistake, this .NET client repo does not currently have a "samples" directory like the Python client does. Should I create that new dir here and place my working sample in that location? Or should I just attach the sample in the same "guides" directory where the MD files live?

dblock:
> @dblock @Xtansia If I'm not mistaken, this .NET client repo does not currently have a "samples" directory like the Python client does. Should I create that new directory here and place my working sample in that location? Or should I just attach the sample in the same "guides" directory where the MD files live?

Please create one!

Djcarrillo6:
@dblock I added the new sample directory and added a working sample for my guide. I also change the name of guide file to match what the Python client [guide](https://github.com/opensearch-project/opensearch-py/blob/main/guides/json.md) used.

Djcarrillo6:
Hey @Xtansia,  would you like me to add a link to this guide in the `USER_GUIDE.md` file? If so, which section to you think is best with respect to the guide's link placement.

Xtansia (changes requested):
@Djcarrillo6 A few points:
* Rather than using `new CancellationToken()` you should use `CancellationToken.None`
* The sample code should get added to a new project so that it actually gets compiled
* As far as linking the guide, I think we should add an `Advanced Features` section like some of the other clients have: https://github.com/opensearch-project/opensearch-py/blob/main/USER_GUIDE.md#advanced-features

Djcarrillo6:
> @Djcarrillo6 A few points:
> 
> * Rather than using `new CancellationToken()` you should use `CancellationToken.None`
> * The sample code should get added to a new project so that it actually gets compiled
> * As far as linking the guide, I think we should add an `Advanced Features` section like some of the other clients have: https://github.com/opensearch-project/opensearch-py/blob/main/USER_GUIDE.md#advanced-features

@Xtansia what would adding the sample to a new project entail? Could I just add the sample file to the `src/OpenSearch.Net` directory?

> @Djcarrillo6 A few points:
> 
> * Rather than using `new CancellationToken()` you should use `CancellationToken.None`
> * The sample code should get added to a new project so that it actually gets compiled
> * As far as linking the guide, I think we should add an `Advanced Features` section like some of the other clients have: https://github.com/opensearch-project/opensearch-py/blob/main/USER_GUIDE.md#advanced-features

@Xtansia Can you detail how would I would correctly add this sample to a new project within the repo?

Xtansia:
@Djcarrillo6 You can create and add a new project with the `dotnet` CLI:
```bash
dotnet new console -o samples/Samples
dotnet sln add samples/Samples
```

USER_GUIDE.md
```diff
@@ -303,3 +303,9 @@ var client = new OpenSearchLowLevelClient(config);
 ```
 
 Note the main difference here is that we are instantiating an `OpenSearchLowLevelClient` rather than `OpenSearchClient`, and `ConnectionConfiguration` instead of `ConnectionSettings`.
+
+
+## Advanced Features
+
+- [Making Raw JSON Requests](guides/json.md)
+  - [Sample Code](/samples/Samples/Program.cs)
```
Xtansia, line 311: You do not need the `[Sample Code]` line here. Additionally there should be a entry added to the table of contents at the top of this file for the `Advanced Features`

samples/Samples/Program.cs
```diff
@@ -0,0 +1,49 @@
+﻿using OpenSearch.Client;
```
Xtansia, line 1: This needs the license header added:
```csharp
*
*/
```

samples/Samples/Samples.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
```
Xtansia, line 7: Add a `<IsPackable>False</IsPackable>` here to make sure it won't get accidentally packed for package publishing

samples/Samples/Samples.csproj
```diff
@@ -0,0 +1,14 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="OpenSearch.Client" Version="1.*" />
```
Xtansia, line 11: This should be a project reference rather than a package reference, so that the sample code will be forced to reflect changes in the library.

guides/json.md
```diff
@@ -0,0 +1,52 @@
+- [Making Raw JSON REST Requests](#making-raw-json-rest-requests)
+  - [GET](#get)
+  - [PUT](#put)
+  - [POST](#post)
+  - [DELETE](#delete)
+
+# Making Raw JSON REST Requests
+OpenSearch exposes a REST API that you can use to interact with OpenSearch. The OpenSearch .NET client provides a low-level API that allows you to send raw JSON requests to OpenSearch. This API is useful if you want to use a feature that is not yet supported by the OpenSearch .NET client, but it supported by the OpenSearch REST API.
```
Xtansia, line 8: You can link to the sample code from here, like the python guide does.

Additionally I think it'd make more sense to use the same example actions that the python guide does: https://github.com/opensearch-project/opensearch-py/blob/main/guides/json.md
https://github.com/opensearch-project/opensearch-py/blob/main/samples/json/hello.py

Xtansia (changes requested):

guides/json.md
```diff
@@ -0,0 +1,69 @@
...
+## GET
+The following example returns the server version information via `GET /`.
+
+```csharp
+var versionResponse = await client.LowLevel.DoRequestAsync<DynamicResponse>(HttpMethod.GET, "/", CancellationToken.None);
+Console.WriteLine(versionResponse.Body["version"]["distribution"].ToString(), versionResponse.Body["version"]["number"].ToString()); // Distribution & Version number
+```
+
+# PUT
+The following example creates a document via `PUT /movies/_doc/{id}` with a request body that specifies the document to create.
```
Xtansia, line 19: This and the rest of the descriptions are no-longer accurate

Xtansia (changes requested):

guides/json.md
```diff
@@ -0,0 +1,69 @@
...
+var indexBody = new 
+{
+    settings = new 
+    {
+        index = new 
+        {
+            number_of_shards = 4
+        }
+    }
+};
```
dblock, line 31: Is there a JSON/string version of this? Not sure what's best for this example, maybe both?

dblock (commented):
Let's make a subfolder for the sample and call it "RawJsonSample"?

Djcarrillo6:
@Xtansia I added the new folder structure & converted the sample to use raw JSON to match the guide implementation as @dblock requested.

Xtansia (changes requested):
If you're using a string body, you should use `PostData.String` rather than `PostData.Serializable`. The guide examples need to be updated to match the sample code. It'd be good to add a section to the guide informing users about the different types of `PostData` they can use with examples.

Djcarrillo6:
> If you're using a string body, you should use `PostData.String` rather than `PostData.Serializable`. The guide examples need to be updated to match the sample code. It'd be good to add a section to the guide informing users about the different types of `PostData` they can use with examples.

@Xtansia I made the change to PostData.String and added a section to the guide informing users about the different `PostData` they can use with examples. Let me know if you'd like anything changed

Xtansia (approved):

dblock:
Looking at how this is done in the .NET client, would aliasing `LowLevel` to `.http` make sense to make the API more digestable? Similar to https://github.com/opensearch-project/opensearch-py/pull/544.

PR #447, opened 2023-11-26
Add higher-level HTTP DSL methods

### LINKED ISSUE #403: [FEATURE] Make raw JSON REST requests to OpenSearch

Xtansia:
### Description
Adds the higher-level HTTP DSL methods to wrap `DoRequest` and allow better ergonomics.

### Issues Resolved
Closes #403 


dblock (commented):
Why do we need `.LowLevel` at all?

Xtansia:
> Why do we need `.LowLevel` at all?

All requests are eventually handled via `.LowLevel`/`OpenSearch.Net`. However in future it'd be ideal to simplify and only have a single client package keeping just the transport layer separate, similar to what NEST did in v8.

dblock (approved):
Looks OK. I think I'd remove any mention of `LowLevel` from the documentation. In my opinion writing `client.Http.GetAsync` is 100x better.

Xtansia:
Converting to draft as there's some naming conventions that need fixing or excluding from conventions tests

dblock (approved):
Retry tests till they pass? 😢

PR #489, opened 2023-12-25
Add support for `PATCH` in `HttpConnection.ConvertHttpMethod`

### LINKED ISSUE #488: [BUG] Missing Case for `HttpMethod.PATCH` in `OpenSearch.Net.HttpConnection.ConvertHttpMethod`
### Bug Description

There is a bug in the `OpenSearch.Net.HttpConnection` implementation of `OpenSearch.Net.IConnection`, specifically in the `ConvertHttpMethod` method. The method is missing a case for handling `HttpMethod.PATCH`, resulting in an exception when this method is used.

### Steps to Reproduce
1. Upgrade to OpenSearch.NET 1.6.0.
2. Create an instance of `OpenSearchClient` using `HttpConnection` or `AwsSigV4HttpConnection` in the `ConnectionSettings` parameter of the client's constructor.
3. Use the `IOpenSearchLowLevelClient.DoRequestAsync` method with `HttpMethod.PATCH`.
4. An exception is thrown, indicating an "Invalid value for HttpMethod PATCH".

### Expected Behavior
The `PATCH` `HttpMethod`, which is now part of the `HttpMethod` enum in version 1.6.0, should be correctly handled by the `ConvertHttpMethod` method without throwing an exception.

### Actual Behavior
The `ConvertHttpMethod` method does not include a case for `HttpMethod.PATCH`, leading to an exception when this HTTP method is used.

### Suggested Fix
Add the missing case for `HttpMethod.PATCH` in the `ConvertHttpMethod` method to align with the updated `HttpMethod` enum.

assaftir:
### Description
Add support for `PATCH` in `HttpConnection.ConvertHttpMethod`.

### Issues Resolved
Fixes #488 

### DIFF
--- a/tests/Tests/Connection/Http/HttpConnectionTests.cs
+++ b/tests/Tests/Connection/Http/HttpConnectionTests.cs
@@ -1,0 +1,52 @@
+/* SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+using System;
+using System.Net.Http;
+using FluentAssertions;
+using OpenSearch.Net;
+using OpenSearch.OpenSearch.Xunit.XunitPlumbing;
+using Xunit;
+using HttpMethod = OpenSearch.Net.HttpMethod;
+
+namespace Tests.Connection.Http;
+
+public class HttpConnectionTests
+{
+	public static TheoryData<HttpMethod> HttpMethods()
+	{
+		var data = new TheoryData<HttpMethod>();
+		foreach (var httpMethod in Enum.GetValues<HttpMethod>()) data.Add(httpMethod);
+		return data;
+	}
+
+	[TU]
+	[MemberData(nameof(HttpMethods))]
+	public void UsesCorrectHttpMethod(HttpMethod method)
+	{
+		var mockHttpConnection = new MockHttpConnection();
+		mockHttpConnection.Request<StringResponse>(new RequestData(method, "", null, null, null, null));
+		mockHttpConnection.LastRequest.Method.Should().Be(new System.Net.Http.HttpMethod(method.ToString()));
+	}
+
+	private class MockHttpConnection : HttpConnection
+	{
+		public HttpRequestMessage LastRequest { get; private set; }
+
+		protected override HttpRequestMessage CreateHttpRequestMessage(RequestData requestData)
+		{
+			LastRequest = base.CreateHttpRequestMessage(requestData);
+			return LastRequest;
+		}
+
+		public override TResponse Request<TResponse>(RequestData requestData)
+		{
+			CreateHttpRequestMessage(requestData);
+			return new TResponse();
+		}
+	}
+}
--- a/tests/Tests/Connection/Http/HttpConnectionTests.cs
+++ b/tests/Tests/Connection/Http/HttpConnectionTests.cs
@@ -1,15 +1,16 @@
 /* SPDX-License-Identifier: Apache-2.0
- *
- * The OpenSearch Contributors require contributions made to
- * this file be licensed under the Apache-2.0 license or a
- * compatible open source license.
- */
+*
+* The OpenSearch Contributors require contributions made to
+* this file be licensed under the Apache-2.0 license or a
+* compatible open source license.
+*/
 
 using System;
 using System.Net.Http;
 using FluentAssertions;
 using OpenSearch.Net;
 using OpenSearch.OpenSearch.Xunit.XunitPlumbing;
+using Tests.Core.Connection.Http;
 using Xunit;
 using HttpMethod = OpenSearch.Net.HttpMethod;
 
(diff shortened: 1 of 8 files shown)


dblock:
Thanks. Add a test please?

assaftir:
> Thanks. Add a test please?

Sure, will do asap.
Any guidance for expected test?

dblock:
@assaftir No particular guidance, see if there are any existing tests with verbs. If not add something. I'm happy if the code path is exercised and tests would fail without this change.

Xtansia (changes requested):
Hi @assaftir,

Thanks for catching this, don't know how I missed it in the initial feature PR!

I've just got one minor nit about using a const.

As far as testing this, I think the simplest possible would be something like the following added to the `Tests` project:
```csharp
 *
 */

using System;
using FluentAssertions;
using OpenSearch.Net;
using OpenSearch.OpenSearch.Xunit.XunitPlumbing;
using Xunit;

namespace Tests.Connection.Http;

public class HttpConnectionTests
{
	public static TheoryData<HttpMethod> HttpMethods()
	{
		var data = new TheoryData<HttpMethod>();
		foreach (var httpMethod in Enum.GetValues<HttpMethod>()) data.Add(httpMethod);
		return data;
	}

	[TU]
	[MemberData(nameof(HttpMethods))]
	public void ConvertHttpMethod_ConvertsMethod(HttpMethod method)
	{
		var converted = HttpConnection.ConvertHttpMethod(method);
		converted.ToString().Should().BeEquivalentTo(method.ToString());
	}
}
```

But it's debatable whether this is a _good_ test as it's directly prodding an implementation detail of `HttpConnection`. Otherwise could do something like https://github.com/opensearch-project/opensearch-net/blob/main/tests/Tests.Auth.AwsSigV4/AwsSigV4HttpConnectionTests.cs and create a mockable subclass of `HttpConnection`, then using same theory data and calling low-level HTTP method then asserting on the constructed HttpRequestMessage

Xtansia:
This will also need an entry added to the Unreleased section of the CHANGELOG.md under a Fixed heading

assaftir:
> Hi @assaftir,
> 
> 
> 
> Thanks for catching this, don't know how I missed it in the initial feature PR!
> 
> 
> 
> I've just got one minor nit about using a const.
> 
> 
> 
> As far as testing this, I think the simplest possible would be something like the following added to the `Tests` project:
> 
> ```csharp
> 
> /* SPDX-License-Identifier: Apache-2.0
> 
>  *
> 
>  * The OpenSearch Contributors require contributions made to
> 
>  * this file be licensed under the Apache-2.0 license or a
> 
>  * compatible open source license.
> 
>  */
> 
> 
> 
> using System;
> 
> using FluentAssertions;
> 
> using OpenSearch.Net;
> 
> using OpenSearch.OpenSearch.Xunit.XunitPlumbing;
> 
> using Xunit;
> 
> 
> 
> namespace Tests.Connection.Http;
> 
> 
> 
> public class HttpConnectionTests
> 
> {
> 
> 	public static TheoryData<HttpMethod> HttpMethods()
> 
> 	{
> 
> 		var data = new TheoryData<HttpMethod>();
> 
> 		foreach (var httpMethod in Enum.GetValues<HttpMethod>()) data.Add(httpMethod);
> 
> 		return data;
> 
> 	}
> 
> 
> 
> 	[TU]
> 
> 	[MemberData(nameof(HttpMethods))]
> 
> 	public void ConvertHttpMethod_ConvertsMethod(HttpMethod method)
> 
> 	{
> 
> 		var converted = HttpConnection.ConvertHttpMethod(method);
> 
> 		converted.ToString().Should().BeEquivalentTo(method.ToString());
> 
> 	}
> 
> }
> 
> ```
> 
> 
> 
> But it's debatable whether this is a _good_ test as it's directly prodding an implementation detail of `HttpConnection`. Otherwise could do something like https://github.com/opensearch-project/opensearch-net/blob/main/tests/Tests.Auth.AwsSigV4/AwsSigV4HttpConnectionTests.cs and create a mockable subclass of `HttpConnection`, then using same theory data and calling low-level HTTP method then asserting on the constructed HttpRequestMessage

I've implemented a test, let me know if it works, otherwise I'll go the route you took with the 'AwsSigV4' tests.

I'm not a huge fan of unit tests and not familiar with the xUnit framework, so forgive me if I haven't done that right.

tests/Tests/Connection/Http/HttpConnectionTests.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            HttpMethod.GET => System.Net.Http.HttpMethod.Get,
+            HttpMethod.POST => System.Net.Http.HttpMethod.Post,
+            HttpMethod.PUT => System.Net.Http.HttpMethod.Put,
+            HttpMethod.DELETE => System.Net.Http.HttpMethod.Delete,
+            HttpMethod.HEAD => System.Net.Http.HttpMethod.Head,
+            HttpMethod.PATCH => Patch,
+            _ => throw new ArgumentException("Invalid value for HttpMethod", nameof(httpMethod))
+        };
+
```
Xtansia, line 60: Rather than duplicating the switch statement here, it'd be better to simplify the test side of this to something like:
```suggestion
		mockHttpConnection.LastRequest.Method.Should().Be(new System.Net.Http.HttpMethod(method.ToString()));
    }
```

tests/Tests/Connection/Http/HttpConnectionTests.cs
```diff
@@ -0,0 +1,61 @@
+using System;
```
Xtansia, line 1: Please add the license header here:
```
 *
 */
```

Xtansia (changes requested):

Xtansia (dismissed):

Xtansia (approved):

Issue #375, opened 2023-09-25
[FEATURE] Provide a way to make `.Strict(true)` be the default behavior for all queries


DumboJet:
### Is your feature request related to a problem?
Applying the `.Strict(true)` to each sub-query is a lot of work and you can always miss something.

### What solution would you like?
Having a way to make `.Strict(true)` be the default behavior (perhaps for a `QueryContainer` and all its children?) would probably be a nice feature. It would save people from a lot of silent bugs, where queries become conditionless and are not serialized.

PR #503, opened 2024-01-08
Fix for #380.

### LINKED ISSUE #380: [BUG] `IEnumerable<int?>` get mapped incorrectly.
### What is the bug?
When your document contains a field of type `IEnumerable<int?>` (or any other `IEnumerable<Nullable<T>>` apparently) it gets mapped incorrectly when auto-mapping. Instead of an integer type, an object type gets created, that contains the nullable properties.

When mapping a nullable object property like `public int? Value { get; set; }` the mapper correctly maps this field into an integer.

But when you map an array of nullable objects like `public IEnumerable<int?> Values { get; set; }` then it maps the nullable class members instead, like this:
image

A mapping for an integer gets created instead.

Windows 11
OpenSearch.Client nuget version 1.5.0
OpenSearch Version: 2.9.0
OpenSearch Security Version: 2.9.0.0

DumboJetEngine:
### Description
Fix for #380.

### Issues Resolved
#380.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Removed
 - Removed the `Features` API which is not supported by OpenSearch from the low-level client ([#331](https://github.com/opensearch-project/opensearch-net/pull/331))
 
+
+### Fixed
+- Fixed `IEnumerable<int?>` property mapping. ([#503](https://github.com/opensearch-project/opensearch-net/pull/503))
+
+
 ### Dependencies
 - Bumps `Microsoft.CodeAnalysis.CSharp` from 4.2.0 to 4.6.0
 - Bumps `NSwag.Core.Yaml` from 13.19.0 to 13.20.0
--- a/src/OpenSearch.Client/Mapping/Visitor/PropertyWalker.cs
+++ b/src/OpenSearch.Client/Mapping/Visitor/PropertyWalker.cs
@@ -205,11 +205,25 @@ namespace OpenSearch.Client
 			if (type.IsArray)
 				return type.GetElementType();
 
-			if (type.IsGenericType && type.GetGenericArguments().Length == 1
-				&& (type.GetInterfaces().HasAny(t => t == typeof(IEnumerable)) || Nullable.GetUnderlyingType(type) != null))
-				return type.GetGenericArguments()[0];
+			if (ShouldUnwrapType(type))
+			{
+				var returnType = type.GetGenericArguments()[0];
+				if (ShouldUnwrapType(returnType)) // This is needed for types like IEnumerable<int?>.
+				{
+					return returnType.GetGenericArguments()[0];
+				}
+				return returnType;
+			}
 
 			return type;
 		}
+
+		private static bool ShouldUnwrapType(Type ty) =>
+					ty.IsGenericType &&
+					ty.GetGenericArguments().Length == 1 &&
+					(
+						ty.GetInterfaces().HasAny(t => t == typeof(IEnumerable)) ||
+						Nullable.GetUnderlyingType(ty) != null
+					);
 	}
 }
--- a/tests/Tests/Mapping/PropertyWalkerTests.cs
+++ b/tests/Tests/Mapping/PropertyWalkerTests.cs
@@ -1,0 +1,84 @@
+/* SPDX-License-Identifier: Apache-2.0
+*
+* The OpenSearch Contributors require contributions made to
+* this file be licensed under the Apache-2.0 license or a
+* compatible open source license.
+*/
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using FluentAssertions;
+using OpenSearch.Client;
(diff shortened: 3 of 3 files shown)


src/OpenSearch.Client/Mapping/Visitor/PropertyWalker.cs
```diff
@@ -205,11 +205,27 @@ private static Type GetUnderlyingType(Type type)
...
+				if (ShouldUnwrapType(returnType)) // This is needed for types like IEnumerable<int?>.
+				{
+					return returnType.GetGenericArguments()[0];
+				}
+				return returnType;
+			}
 
 			return type;
+
+			bool ShouldUnwrapType(Type ty)
```
dblock, line 220: Is this a pattern elsewhere in the codebase to write a function within another function? Otherwise make this `private static`?

dblock (commented):
Thanks. Will need a unit test as well.

src/OpenSearch.Client/Mapping/Visitor/PropertyWalker.cs
DumboJetEngine: Apparently not.
Although, it might be because local functions are a relatively new language feature.
I can change this.

For the test, I saw that there was no tests at all for the `PropertyWalker` and the `PropertiesExtensions` that was using it was marked as internal. I will add some tests now.

DumboJetEngine: I guess I also need to update the change-log too?
Anything special I should pay attention to?

dblock: Nope, keep it short! :) Thanks much.

DumboJetEngine:
> Thanks. Will need a unit test as well.

Shall I put it at `Tests.Reproduce.GitHubIssue380` or this is some code area that is inherited by ElasticSearch?
I see it contains 4-digit issues in there, while OpenSearch is still using 3-digit issues. That's why I ask -- it feels weird.

DumboJetEngine:
I have mixed the two issues ( #380, #375 ) a bit and messed up the history. 🤦‍♂️ 
Shall I perhaps create the PR from scratch?

Xtansia:
> I have mixed the two issues ( #380, #375 ) a bit and messed up the history. 🤦‍♂️ Shall I perhaps create the PR from scratch? Other than that, I think the changes should be OK now.
> 
> Update: Perhaps I should move my tests inside the `Tests` project, and out of `Tests.Reproduce`.

@DumboJetEngine You don't need to completely recreate the PR, you can amend and drop the commits on your local branch as you see fit and then force push your feature branch and the content of the PR will update to match. This is preferred over closing and recreating the PR, as otherwise it can create a lot of notification spam and churn in the closed PR history.

DumboJetEngine:
@Xtansia 
OK.
Done. :)
Let me know if all is OK.

CHANGELOG.md
```diff
@@ -14,6 +14,11 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Removed
 - Removed the `Features` API which is not supported by OpenSearch from the low-level client ([#331](https://github.com/opensearch-project/opensearch-net/pull/331))
 
+
+### Fixed
+- Fixed `IEnumerable<int?>` property mapping. ([#380](https://github.com/opensearch-project/opensearch-net/pull/375))
```
Xtansia, line 19: The changelog item should reference this PR's number: 503
```suggestion
- Fixed `IEnumerable<int?>` property mapping. ([#503](https://github.com/opensearch-project/opensearch-net/pull/503))
```

Xtansia (changes requested):
Thanks @DumboJetEngine, code-wise all looks good to me, just need to correct the PR number in the changelog item

tests/Tests/Mapping/PropertyWalkerTests .cs
```diff
@@ -0,0 +1,77 @@
+using System;
```
Xtansia, line 1: This file needs a license header like so:
```suggestion
/* SPDX-License-Identifier: Apache-2.0
*
* The OpenSearch Contributors require contributions made to
* this file be licensed under the Apache-2.0 license or a
* compatible open source license.
*/
using System;
```

Xtansia (changes requested):
Sorry just noticed a couple other things, the test file needs a license header added, and it appears you have a space in the file name of the test file.

tests/Tests/Mapping/PropertyWalkerTests .cs
DumboJetEngine: Fixed. Anything more I need to change?

Xtansia (approved):

Xtansia:
@DumboJetEngine Would you be able to follow the above instructions fixing any merge conflicts (almost certainly only something minor in the CHANGELOG) to backport this fix to the 1.x version line?

DumboJetEngine:
@Xtansia Yes, I think I can do this.
Tomorrow though.
I need to leave my computer now.

PR #504, opened 2024-01-08
fix for #375

### LINKED ISSUE #375: [FEATURE] Provide a way to make `.Strict(true)` be the default behavior for all queries

DumboJetEngine:
### Description
fix for #375

### Issues Resolved
#375


dblock:
Thanks @DumboJetEngine! Write a test please? Also in your commits, describe what is fixed rather than just a fix for a bug number.

PR #507, opened 2024-01-09
[Backport 1.x] Add support for `PATCH` in `HttpConnection.ConvertHttpMethod` (#489)

Xtansia:
### Description
Backports 4e721ee854f90fb66f05002c991492c081b68bf0 from #489 

### DIFF
--- a/src/OpenSearch.Net/Connection/HttpConnection.cs
+++ b/src/OpenSearch.Net/Connection/HttpConnection.cs
@@ -28,7 +28,6 @@
 
 #if DOTNETCORE
 using System;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Diagnostics;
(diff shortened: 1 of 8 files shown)


src/OpenSearch.Net/Connection/HttpConnection.cs
```diff
@@ -431,19 +432,17 @@ private static async Task SetContentAsync(HttpRequestMessage message, RequestDat
...
-				default:
-					throw new ArgumentException("Invalid value for HttpMethod", nameof(httpMethod));
-			}
-		}
+				HttpMethod.GET => System.Net.Http.HttpMethod.Get,
+				HttpMethod.POST => System.Net.Http.HttpMethod.Post,
+				HttpMethod.PUT => System.Net.Http.HttpMethod.Put,
+				HttpMethod.DELETE => System.Net.Http.HttpMethod.Delete,
+				HttpMethod.HEAD => System.Net.Http.HttpMethod.Head,
+				HttpMethod.PATCH => Patch,
```
dblock, line 443: I think if I delete this line the tests still pass because they use this list to iterate over the methods being checked. I think you should duplicate it in tests.

Xtansia: The test iterates based on the declared values of the libraries HttpMethod enum, not based on this switch. Which covers the bug that was introduced, Patch was added to the enum, but this didn't have a case added and fell through to the default exception. The test would have automatically added the PATCH test case based on the enum, and then subsequently failed.

dblock: Oh my bad, I see.

dblock (approved):

PR #509, opened 2024-01-11
Fix for #375 - Added `.Strict(bool)` and `.Verbatim(bool)` methods on…

### LINKED ISSUE #375: [FEATURE] Provide a way to make `.Strict(true)` be the default behavior for all queries

DumboJetEngine:
### Description
Fix for #375 - Added `.Strict(bool)` and `.Verbatim(bool)` methods on `QueryContainer` that recursively apply the respective attribute to all subqueries. Also added a `.Name(string)` method on `QueryContainer` that names the (root) contained query.

### Issues Resolved
#375

### DIFF
--- a/src/OpenSearch.Client/QueryDsl/Abstractions/Container/QueryContainer-Dsl.cs
+++ b/src/OpenSearch.Client/QueryDsl/Abstractions/Container/QueryContainer-Dsl.cs
@@ -131,5 +131,58 @@ namespace OpenSearch.Client
 
 		// ReSharper disable once UnusedMember.Global
 		internal bool ShouldSerialize(IJsonFormatterResolver formatterResolver) => IsWritable;
+
+		/// <summary>
+		/// Assigns a name to the contained query.
+		/// </summary>
+		/// <param name="name"></param>
+		/// <returns></returns>
+		public QueryContainer Name(string name)
+		{
+			ContainedQuery.Name = name;
+			return this;
+		}
+
+		/// <summary>
+		/// Applies or removes the `strict` attribute to the contained query and optionally to all child sub-queries.
+		/// </summary>
+		/// <param name="strict"></param>
+		/// <param name="recurse">When true, it applies the attribute to all child sub-queries.</param>
+		/// <returns></returns>
+		public QueryContainer Strict(bool strict = true, bool recurse = false)
+		{
+			if (recurse)
+			{
+				var visitor = new StrictnessPropagatingVisitor(strict);
+				Accept(visitor);
+			}
+			else
+			{
+				ContainedQuery.IsStrict = strict;
+			}
+
+			return this;
+		}
+
+		/// <summary>
+		/// Applies or removes the `verbatim` attribute to the contained query and optionally to all child sub-queries.
+		/// </summary>
+		/// <param name="verbatim"></param>
+		/// <param name="recurse">When true, it applies the attribute to all child sub-queries.</param>
+		/// <returns></returns>
+		public QueryContainer Verbatim(bool verbatim = true, bool recurse = false)
+		{
+			if (recurse)
+			{
+				var visitor = new VerbatimPropagatingVisitor(verbatim);
+				Accept(visitor);
+			}
+			else
+			{
+				ContainedQuery.IsVerbatim = verbatim;
+			}
+
+			return this;
+		}
 	}
 }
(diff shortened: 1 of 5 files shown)


src/OpenSearch.Client/QueryDsl/Abstractions/Container/QueryContainer-Dsl.cs
```diff
@@ -131,5 +131,42 @@ public void Accept(IQueryVisitor visitor)
...
+		public QueryContainer Name(string name)
+		{
+			ContainedQuery.Name = name;
+			return this;
+		}
+
+		/// <summary>
+		/// Applies or removes the `strict` attribute to the query container and all child sub-queries.
+		/// </summary>
+		/// <param name="strict">When null, the value is taken from the root query.</param>
```
Xtansia, line 145: This "When null, .." behaviour feels somewhat unintuitive for what a null value should do and I'm unsure of the use-case. It might be better to keep that as something that's done intentionally via something like `container.Strict(container.IsStrict, recurse: true)`

src/OpenSearch.Client/QueryDsl/Abstractions/Container/QueryContainer-Dsl.cs
```diff
@@ -131,5 +131,42 @@ public void Accept(IQueryVisitor visitor)
...
+		/// <returns></returns>
+		public QueryContainer Strict(bool? strict = true)
```
Xtansia, line 147: I'd prefer if this was something like `Strict(bool strict = true, bool recurse = false)` so that the recurse is something opted into explicitly, and otherwise behaves the same as all other strict setters.

src/OpenSearch.Client/QueryDsl/Abstractions/Container/QueryContainer-Dsl.cs
```diff
@@ -131,5 +131,42 @@ public void Accept(IQueryVisitor visitor)
...
+		{
+			if (!strict.HasValue)
+				strict = ContainedQuery.IsStrict;
+
+			var visitor = new QueryNodeModifierVisitor((query, ctx) => query.IsStrict = strict.Value);
```
Xtansia, line 152: It may end up being simpler and less code to just make two very short specialised implementations of QueryVisitor that can be class private ie something like `new StrictnessPropogatingVisitor(bool strict)`

Xtansia (changes requested):
Hi @DumboJetEngine,

Apologies for taking a bit to getting to review this.

I've added a few comments with my thoughts.

src/OpenSearch.Client/QueryDsl/Visitor/QueryNodeModifierVisitor.cs
```diff
@@ -0,0 +1,32 @@
...
+* The OpenSearch Contributors require contributions made to
+* this file be licensed under the Apache-2.0 license or a
+* compatible open source license.
+*/
+
+using System;
+
+namespace OpenSearch.Client
+{
+	public class QueryNodeModifierVisitor : QueryVisitor
```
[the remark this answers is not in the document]
DumboJetEngine: Yes, I was thinking about that too.
But I wasn't creative enough to figure out a better name at that point.
I will try again.

src/OpenSearch.Client/QueryDsl/Abstractions/Container/QueryContainer-Dsl.cs
DumboJetEngine: Yes, I was initially going for a method signature of `Strict(bool isStrict)`, but then some tests requiring special method signature for single-argument methods on `IDescriptor` started failing. I am not sure why those requirements exist. But your suggestion seems nice.

src/OpenSearch.Client/QueryDsl/Abstractions/Container/QueryContainer-Dsl.cs
DumboJetEngine: I usually go for more general-purpose implementations, but your suggestion will definitely solve my problem finding a good name for my current class implementation... ;)

src/OpenSearch.Client/QueryDsl/Abstractions/Container/QueryContainer-Dsl.cs
Xtansia: I think I know what tests you're talking about, if you just ignore that test for now, then we can figure out the best way to have the test ignore this afterwards

Xtansia (approved):

Xtansia:
Thanks @DumboJetEngine!

PR #568, opened 2024-02-29
Support list for `opensearch_version` in PluginStats

Xtansia:
### Description
Support list for `opensearch_version` in PluginStats. Fixes failing tests against unreleased OpenSearch `2.x` & `main`.

### DIFF
--- a/.github/workflows/integration-yaml-tests.yml
+++ b/.github/workflows/integration-yaml-tests.yml
@@ -14,6 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
+          - 2.12.0
           - 2.11.1
           - 2.10.0
           - 2.8.0
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -17,6 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
+          - 2.12.0
           - 2.11.1
           - 2.10.0
           - 2.8.0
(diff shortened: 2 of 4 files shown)


dblock (approved):

dblock (changes requested):
Is there a breaking change being introduced in 2.13? That shouldn't be the case.

Xtansia:
> Is there a breaking change being introduced in 2.13? That shouldn't be the case.

It's a result of this backport: https://github.com/opensearch-project/OpenSearch/pull/12271

dblock:
It's a bug, I opened https://github.com/opensearch-project/OpenSearch/issues/12528. Can we avoid this breakage in a different way (mark it pending?).

Xtansia:
> It's a bug, I opened [opensearch-project/OpenSearch#12528](https://github.com/opensearch-project/OpenSearch/issues/12528). Can we avoid this breakage in a different way (mark it pending?).

What exactly do you mean by "mark it pending"? Change the PR to Draft? As presumably we'll still need these changes as the breakage will remain in 3.x?

dblock:
> What exactly do you mean by "mark it pending"? Change the PR to Draft? As presumably we'll still need these changes as the breakage will remain in 3.x?

I mean our build is broken here without this change and we need a workaround, not support for an array. I believe that was a side-effect/mistake and we will not be making a breaking change in 3.0, but I could be wrong.

PR #609, opened 2024-04-16
Update API generator to use new OpenAPI specification

### LINKED ISSUE #580: Update code generator to use new version of OpenAPI specification
The https://github.com/opensearch-project/opensearch-api-specification repository has been updated to be written natively in OpenAPI, rather than being built from Smithy. This means that some structure and details of the specification has changed. The code generator should be updated to use this new version of the specification. Any changes to the generated output should be validated to confirm if the changes are correct by cross referencing with the OpenSearch implementation & documentation. Any errors found in the specification should be reported on the specification repository.

Xtansia:
### Description
Update API generator to use new OpenAPI specification

Fixes #580 


src/OpenSearch.Net/Api/Enums.cs
```diff
@@ -148,17 +148,6 @@ public enum Size
...
-	[StringEnum]
-	public enum Level
-	{
-		[EnumMember(Value = "cluster")]
-		Cluster,
-		[EnumMember(Value = "indices")]
-		Indices,
-		[EnumMember(Value = "shards")]
-		Shards
-	}
```
dblock, line 160: Is this a breaking API change?

src/OpenSearch.Net/_Generated/Api/RequestParameters/RequestParameters.Cluster.cs
```diff
@@ -281,13 +300,6 @@ public string AwarenessAttribute
             set => Q("awareness_attribute", value);
         }
 
-        /// <summary>Specify the level of detail for returned information.</summary>
```
dblock, line 284: Another potentially breaking change.

dblock (changes requested):
There are a few changes here that looks breaking API-wise, e.g. `HealthStatus` replaces `WaitForStatus`, etc. 

Either bump major version in this PR and add UPGRADING.md, or add backwards compatible code?

Xtansia:
> There are a few changes here that looks breaking API-wise, e.g. `HealthStatus` replaces `WaitForStatus`, etc.
> 
> Either bump major version in this PR and add UPGRADING.md, or add backwards compatible code?

`main` is already developing the next major version (2.x) and includes other breaking changes. I'll make an UPGRADING.md

dblock (approved):

PR #623, opened 2024-04-30
Add support for MinScore in ScriptScoreQuery

SamuelCox:
### Description
Add support for `min_score` property in ScriptScoreQuery

### Issues Resolved
Closes #622 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

Issue #282, opened 2023-07-16
[BUG] running `./build.sh` on `main` has failing tests


david-alpert-nl:
### What is the bug?
the `main` branch of this repo has failing tests.

### How can one reproduce the bug?

- clone this repo
- check out the `main` branch
- run `./build.sh`

### What is the expected behavior?
- the build and all tests run as part of the build pass, creating a stable baseline against which contributors can submit new changes

### What is your host/environment?

- OS: macOS Monterey v12.6
- MacBook Pro (16-inch, 2021)
- Chip: Apple M1 Max

### Do you have any screenshots?
![image](https://github.com/opensearch-project/opensearch-net/assets/112503249/aea20e47-9ef3-4b19-8887-2e419fcd5e16)

### Do you have any additional context?
Having a suite of unit tests is very beneficial to new contributors to demonstrate a stable baseline against any new changes.  When that test suite is not maintained it creates a higher barrier to contributors than if there are no tests at all.

Xtansia:
There are a handful of flaky tests that haven't (yet) received any in depth investigation to confirm why they're failing.

It'd be useful if you could share the full output from running `./build.sh` to confirm if this is just those flaky tests or some other issue.

david-alpert-nl:
@Xtansia as a workaround the command you provided in GH-283 against `main` did work for me to complete a test run without any failing tests as a useful baseline:
```
./build.sh seed:6572 random:sourceserializer:false random:typedkeys:false random:httpcompression test "andoperator,notoperator,oroperator,unaryaddoperator,booldslcomplexquery,prefixquery,termquery,termquerywithcaseinsensitive,wildcardquery,strictquery"

...

:: test: Succeeded. (7.64 s)
:: ─────────────────────────────────────
:: Duration        Outcome    Target
:: ──────────────  ─────────  ──────────
:: 2 ms     0.0%   Succeeded  clean
:: 15 ms    0.1%   Succeeded  version
:: 1.90 s   8.9%   Succeeded  restore
:: 7.64 s   36.0%  Succeeded  test
:: 11.68 s  55.0%  Succeeded  full-build
:: ─────────────────────────────────────
:: Succeeded. (test) (21.2 s)
```

david-alpert-nl:
output from running `./build.sh` on `main`
```
./build.sh

false exist:: Starting... (build)
:: clean: Starting...
SKIPPED target 'clean' evaluated not to run
:: clean: Succeeded. (2.62 ms)
[...]

PR #624, opened 2024-04-30
feat: add support for min_score on ScriptScoreQuery

### LINKED ISSUE #622: [BUG] No MinScore property exposed on ScriptScore query
### What is the bug?
The docs specify an optional MinScore property https://opensearch.org/docs/latest/query-dsl/specialized/script-score/

Unless I am missing something, there is no method to specify this in opensearch.client

``` var kNNQueryDescriptor = new ScriptScoreQueryDescriptor<Product>();

 if (searchCriteria.Embedding == null || searchCriteria.Embedding.Length == 0)
 {
     return kNNQueryDescriptor;
 }

 kNNQueryDescriptor
     .Query(q => q.Bool(x => x.Must(y => y.Term(f => f.SoldOut, false))))
     .Script(s => s.Source("knn_score")
                   .Lang("knn")
                   
                   .Params( new Dictionary<string, object> {
                       { "field", "searchableTextEmbedding" }, 
                       { "query_value", searchCriteria.Embedding }, 
                       { "space_type", "cosinesimil" } 
                   }));

 return kNNQueryDescriptor;
```
 
 Where in above can I set the MinScore property?

_A clear and concise description of what you expected to happen._

Windows 11

### Do you have any screenshots?
image
image
image

_Add any other context about the problem._

SamuelCox:
### Description
Add support for `min_score` property in `ScriptScoreQuery`. This also is a proper version of https://github.com/opensearch-project/opensearch-net/pull/623 which I closed after realising trying to add something to an SDK in the github browser editor was probably a bad idea.

Also, when running `build.bat` locally I get a few errors, I don't want to take the action minutes up with another duff run, so could someone check if this is just something not correctly set up on my machine, or an actual error I've caused?
image
image
I could be missing it, but I couldn't find a guide on how to run these tests so I'm assuming I'm missing some setup

### Issues Resolved
Closes #622 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,6 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Added
 - Added support for `MinScore` on `ScriptScoreQuery` ([#624](https://github.com/opensearch-project/opensearch-net/pull/624))
-
 - Added support for the `Cat.PitSegments` and `Cat.SegmentReplication` APIs ([#527](https://github.com/opensearch-project/opensearch-net/pull/527))
 
 ### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - As part of [efforts to re-generate the client](https://github.com/opensearch-project/opensearch-net/pulls?q=is%3Apr+label%3Acode-gen+is%3Aclosed) from our [OpenAPI specification](https://github.com/opensearch-project/opensearch-api-specification) there have been numerous corrections and changes that resulted in breaking changes. Please refer to [UPGRADING.md](UPGRADING.md) for a complete list of these breakages and any relevant guidance for upgrading to this version of the client.
 
 ### Added
+- Added support for `MinScore` on `ScriptScoreQuery` ([#624](https://github.com/opensearch-project/opensearch-net/pull/624))
+
 - Added support for the `Cat.PitSegments` and `Cat.SegmentReplication` APIs ([#527](https://github.com/opensearch-project/opensearch-net/pull/527))
 
 ### Removed
--- a/src/OpenSearch.Client/QueryDsl/Specialized/ScriptScore/ScriptScoreQuery.cs
+++ b/src/OpenSearch.Client/QueryDsl/Specialized/ScriptScore/ScriptScoreQuery.cs
@@ -52,6 +52,12 @@ namespace OpenSearch.Client
 		/// </summary>
 		[DataMember(Name = "script")]
 		IScript Script { get; set; }
+
+        /// <summary>
+		/// The score above which documents will be returned
+		/// </summary>
+		[DataMember(Name = "min_score")]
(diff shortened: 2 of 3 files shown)


SamuelCox:
I can mark this as ready for review if the maintainers are happy, I just thought I should wait to figure out why those tests are failing first.

SamuelCox:
Have marked as ready for review so it gets in front of maintainers, but like I say I'm not sure what's causing those test failures on my local machine

SamuelCox:
image
Seems like those NullRefs are gone after trying a couple more times, so I guess they were just flaky.

dblock:
@tfarrep where does this fall in the code generation?

@SamuelCox check whether this field exists in https://github.com/opensearch-project/opensearch-api-specification, please and contribute there if it's not, we're trying to get most code generated from spec

@SamuelCox on the errors, I think we need a developer guide like https://github.com/opensearch-project/opensearch-py/blob/main/DEVELOPER_GUIDE.md (want to start one?) and we can debug what's going on with your own setup

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - As part of [efforts to re-generate the client](https://github.com/opensearch-project/opensearch-net/pulls?q=is%3Apr+label%3Acode-gen+is%3Aclosed) from our [OpenAPI specification](https://github.com/opensearch-project/opensearch-api-specification) there have been numerous corrections and changes that resulted in breaking changes. Please refer to [UPGRADING.md](UPGRADING.md) for a complete list of these breakages and any relevant guidance for upgrading to this version of the client.
 
 ### Added
+- Added support for `MinScore` on `ScriptScoreQuery` ([#624](https://github.com/opensearch-project/opensearch-net/pull/624))
+
```
dblock, line 10: Nit: remove empty line.

SamuelCox:
> @tfarrep where does this fall in the code generation?
> 
> 
> 
> @SamuelCox check whether this field exists in https://github.com/opensearch-project/opensearch-api-specification, please and contribute there if it's not, we're trying to get most code generated from spec
> 
> 
> 
> @SamuelCox on the errors, I think we need a developer guide like https://github.com/opensearch-project/opensearch-py/blob/main/DEVELOPER_GUIDE.md (want to start one?) and we can debug what's going on with your own setup
> 
> 
> 
> 
image
It appears to exist in the spec already if I understand correctly. @Xtansia asked me to add it here in the issue I logged but perhaps I misunderstood.

I agree that a dev guide would be good, probably would be best to add as a separate PR though?

Happy to add a follow up when I get time

Xtansia:
@dblock This is a field in the request body and the generator does not yet generate the request/response body structures. 

@SamuelCox Thanks for PR'ing this so quickly, changes look good to me, just waiting on the CI to pass. Unfortunately there is a couple flaky tests floating around, I've made an issue regarding your specific null ref one (#626). The TestEphemeralCluster 403 errors are ones I haven't experienced myself (nor seen in CI), but were also reported in #282. So if you happen to gain any further insight please add it to the relevant issues. I also second that if you wanted to contribute towards a developer's guide that'd be amazing!

Xtansia (approved):

PR #646, opened 2024-05-14
Add interfaces to OpenSearchClient's namespaces API specifications (fixes #423)

### LINKED ISSUE #423: Support mocking and testing of ClusterNamespace
Non-virtual/non-interface implementations of `OpenSearch.Client.Specification.ClusterApi.ClusterNamespace` makes mocking cluster health checks impossible. 

Make methods virtual, or implement interface type for `ClusterNamespace` in `IOpenSearchClient` or make methods in `ClusterNamespace` virtual.

There isn't another good way to do this. Mocking packages should only mock interfaces or public virtual methods.

image

mispencer:
### Description
Say you created a method using IOpenSearchClient:
```
public void CreateIndex(IOpenSearchClient client, string name) {
	client.Indices.Create(name);
}
```

Then you want to write a unit test for it. Since you don't want unit tests to depend on external services, you would likely mock out the IOpenSearchClient.  So, you might write:
```
public void CreateIndex_Always_CreatesAnIndex() {
	// arrange
	var clientMock = new Moq.Mock<IOpenSearchClient>();
	clientMock.Setup(i => i.Indices.Create(name, null));
	var name = "Name";

	// act
	_instance.CreateIndex(client, name)

	// assert
	clientMock.Verify(i => i.Indices.Create(name, null), Times.Once);
}
```

But this doesn't work. IOpenSearchClient is an interface and can be mocked, but IOpenSearchClient.Indices is IndicesNamespace, a concrete class, and concrete classes are not mockable unless they have a public constructor and have visual methods - neither which apply to IndicesNamespace or any other class under OpenSearch.Client.Specification. As such, you get an error on running this test or any test that references or calls code that references any of the namespaced API endpoints.

This PR adds interfaces to all of the auto-generated classes under OpenSearch.Client.Specification, and adjusts IOpenSearchClient and OpenSearchClient to return these interfaces instead of the concrete classes. This allows mocking to work for these classes, so that users of this library can unit test methods using them.

### Issues Resolved
#423 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,9 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ## [Unreleased]
 ### ⚠️ Breaking Changes ⚠️
 - As part of [efforts to re-generate the client](https://github.com/opensearch-project/opensearch-net/pulls?q=is%3Apr+label%3Acode-gen+is%3Aclosed) from our [OpenAPI specification](https://github.com/opensearch-project/opensearch-api-specification) there have been numerous corrections and changes that resulted in breaking changes. Please refer to [UPGRADING.md](UPGRADING.md) for a complete list of these breakages and any relevant guidance for upgrading to this version of the client.
-- To allow easier unit testing of code using OpenSearch, interfaces were introduced for all classes under OpenSearch.Client.Specification, and the properties in IOpenServiceClient and OpenServiceClient were changed from the concrete classes to the new interfaces.
+
+### Changed
+- Changed the namespace client properties on `IOpenSearchClient` to return corresponding interfaces to better enable mocking & unit testing ([#646](https://github.com/opensearch-project/opensearch-net/pull/646))
 
 ### Added
 - Added support for `MinScore` on `ScriptScoreQuery` ([#624](https://github.com/opensearch-project/opensearch-net/pull/624))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ## [Unreleased]
 ### ⚠️ Breaking Changes ⚠️
 - As part of [efforts to re-generate the client](https://github.com/opensearch-project/opensearch-net/pulls?q=is%3Apr+label%3Acode-gen+is%3Aclosed) from our [OpenAPI specification](https://github.com/opensearch-project/opensearch-api-specification) there have been numerous corrections and changes that resulted in breaking changes. Please refer to [UPGRADING.md](UPGRADING.md) for a complete list of these breakages and any relevant guidance for upgrading to this version of the client.
+- To allow easier unit testing of code using OpenSearch, interfaces were introduced for all classes under OpenSearch.Client.Specification, and the properties in IOpenServiceClient and OpenServiceClient were changed from the concrete classes to the new interfaces.
 
 ### Added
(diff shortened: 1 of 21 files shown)


Xtansia (changes requested):
Thank you for making this PR it's greatly appreciated! Just a couple minor things:
- I think it'd be good to also add the interface for the Http namespace for consistency
- You'll need to add an entry to the `CHANGELOG.md` saying this has been added/changed
- Info should also be added to `UPGRADING.md` as changing the property types on the client is technically a breaking change

mispencer:
CHANGELOG and UPGRADING has been adjusted. I am not exactly sure how to layout the UPGRADING change, so feedback on the preferred text and format would be great.

IHttpNamespace has been added.

UPGRADING.md
```diff
@@ -36,6 +39,13 @@
 
 # Upgrading OpenSearch.Net & OpenSearch.Client
 
+## [Unreleased]
+
+### OpenSearch.Client
+
+#### General
+- Namespaced APIs, exposed in `IOpenSearchClient`, have gained an interface, and the properties on `IOpenSearchClient` and `OpenSearchClient` have been changed to the new interfaces. For example, `IOpenSearchClient.Cluster` was `ClusterNamespace` and now is `IClusterNamespace`.
```
Xtansia, line 47: Sorry I should have given more info, `2.0.0` is the in-development unreleased major version, so this can go in the existing `1.x.y to 2.0.0 > OpenSearch.Client > General` section. 

Wording wise I would potentially rephrase this as:
```suggestion
#### General
- The namespaced APIs exposed in `IOpenSearchClient` have each gained a corresponding interface and the types of the properties on `IOpenSearchClient` and `OpenSearchClient` have been changed from the concrete implementations to the matching interfaces. For example, `IOpenSearchClient.Cluster` was `ClusterNamespace` and now is `IClusterNamespace`.
```

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ## [Unreleased]
 ### ⚠️ Breaking Changes ⚠️
 - As part of [efforts to re-generate the client](https://github.com/opensearch-project/opensearch-net/pulls?q=is%3Apr+label%3Acode-gen+is%3Aclosed) from our [OpenAPI specification](https://github.com/opensearch-project/opensearch-api-specification) there have been numerous corrections and changes that resulted in breaking changes. Please refer to [UPGRADING.md](UPGRADING.md) for a complete list of these breakages and any relevant guidance for upgrading to this version of the client.
+- To allow easier unit testing of code using OpenSearch, interfaces were introduced for all classes under OpenSearch.Client.Specification, and the properties in IOpenServiceClient and OpenServiceClient were changed from the concrete classes to the new interfaces.
```
Xtansia, line 7: I think we can move this further down under a `### Changed` heading, and it should generally follow the layout and verbiage of other entries, something like:
```suggestion
- Changed the namespace client properties on `IOpenSearchClient` to return corresponding interfaces to better enable mocking & unit testing ([#646](https://github.com/opensearch-project/opensearch-net/pull/646))
```

Xtansia (changes requested):

UPGRADING.md
mispencer: This has been updated.

CHANGELOG.md
mispencer: This has been updated

Xtansia (approved):
Awesome, thanks @mispencer!

PR #734, opened 2024-07-30
feat(): add support for DateOnly and TimeOnly #680

### LINKED ISSUE #680: [FEATURE] Serialize DateOnly as a date instead of serializing its properties individually
DateOnly properties are treated as any other complex type and its properties are serialised individually. It prevents using features like date ranges when searching.

DateOnly properties be treated as DateTime properties, except maybe they'd use a different format (like `yyyy-mm-dd`)

Customizing the serialization behavior myself, but I didn't find a way to do so. Also tried applying [DateAttribute] to my property but didn't seem to change anything

Currently [DateOnly](https://learn.microsoft.com/en-us/dotnet/api/system.dateonly) properties are serialized this way :
```json
"dateOnlyProperty": {
    "year": 2024,
    "month": 4,
    "day": 15,
    "dayOfWeek": 1,
    "dayOfYear": 106,
    "dayNumber": 738990
},
```

PS : `TimeOnly` may have the same kind of issues

joefeser:
### Description
Add support for DateOnly and TimeOnly that were both added in .net 6

### Issues Resolved
Closes #680.


global.json
```diff
@@ -1,10 +1,10 @@
 {
   "sdk": {
-    "version": "6.0.403",
+    "version": "8.0.107",
```
Xtansia, line 3: The github workflow `setup-dotnet` steps can be updated to change the sdk versions being installed.

src/OpenSearch.Net/Transport/PostData.cs
```diff
@@ -85,8 +85,8 @@ public abstract class PostData
 
 		public static PostData Bytes(byte[] bytes) => new PostData<object>(bytes);
 
-#if NETSTANDARD2_1
-		public static PostData ReadOnlyMemory(ReadOnlyMemory<byte> bytes) => new PostData<object>(bytes);
+#if NETSTANDARD2_1 || NET6_0_OR_GREATER
+		public static PostData ReadOnlyMemory(ReadOnlyMemory<byte> bytes) => new PostData<object>(bytes.ToArray());
```
Xtansia, line 89: Please remove the `.ToArray()`, this should be using the constructor here: https://github.com/opensearch-project/opensearch-net/blob/main/src/OpenSearch.Net/Transport/PostData.cs#L149

You'll just need to update the `#if`s further down.

tests/Tests/ClientConcepts/LowLevel/PostData.doc.cs
```diff
@@ -1,3 +1,5 @@
+#pragma warning disable IDE1006 // Naming Styles
+#pragma warning disable IDE0044 // Add readonly modifier
```
Xtansia, line 2: Please move these `#pragma`s below the license header

Xtansia (changes requested):
Thanks for tackling this @joefeser!
Some relatively small issues other than those already commented in particular files:
- Looks like there's a compile error with the change in SDK.
- Please add an entry to the CHANGELOG.md under the `### Added` section
- The DCO check requires commits to be signed off, you can do this for your existing commits via `git rebase HEAD~2 --signoff` and then make sure to pass `--signoff` to any future commits for this repo

joefeser:
Working on the changes as we speak. Thank you for being so proactive on my PR

global.json
joefeser: Resolved

tests/Tests/ClientConcepts/LowLevel/PostData.doc.cs
joefeser: Resolved

src/OpenSearch.Net/Transport/PostData.cs
joefeser: I think I understood this comment, it is in this commit bcbd1a2161d263ecb010daf522896965ba39531f

joefeser:
> Thanks for tackling this @joefeser! Some relatively small issues other than those already commented in particular files:
> 
> * Looks like there's a compile error with the change in SDK.
> * Please add an entry to the CHANGELOG.md under the `### Added` section
> * The DCO check requires commits to be signed off, you can do this for your existing commits via `git rebase HEAD~2 --signoff` and then make sure to pass `--signoff` to any future commits for this repo

All of these tasks are completed

Xtansia (approved):

Xtansia:
Thank you for this @joefeser, would you mind making a follow up PR adding the DateOnly and TimeOnly types into the unit test here: https://github.com/opensearch-project/opensearch-net/blob/main/tests/Tests.Reproduce/DateSerialization.cs

joefeser:
@Xtansia I will add it in the next few days. I wasn't sure where to write the tests. I am going to do a POC also to add NodaTime support and at least get that conversation going

nalka0:
I have played a bit with this PR and I think there we still have some likely problems :
- If you don't do anything to the mappings, your `TimeOnly` fields are mapped as strings with
```json
"type": "text",
"fields": {
    "keyword": {
        "type": "keyword",
        "ignore_above": 256
    }
}
```
instead of something with `"type": "date"`
- If you use `openSearchClient.Indices.CreateAsync(indexName, x => x.Map(x => x.AutoMap(typeof(YourType))))` your `DateOnly` and `TimeOnly` properties have no way to be inferred and will be still be treated as "normal" objects here. https://github.com/opensearch-project/opensearch-net/blob/c2fdd5d1df4929000249d07ccdf2c8b21fd2fde6/src/OpenSearch.Client/Mapping/Visitor/PropertyWalker.cs#L197

I tried fixing this second issue and for `DateOnly`, it seems like adding a `case "DateOnly"` next to the DateTime one above on the method I linked is sufficient.

On the other hand for `TimeOnly` it seems like we need more than that since you get a parsing error when trying to put data in your index. From my understanding, the format is responsible there but adding
```csharp
case "TimeOnly":
    return new DateProperty()
    {
        Format = "strict_hour_minute_second_millis"
    };
```
was unsuccesful (for some reasons, doing `Format = ...` there doesn't seem to change anything to the resulting mappings).

PS : I pulled the value for `Format = ` [here](https://opensearch.org/docs/latest/field-types/supported-field-types/date/#formats) but am unsure if `strict_hour_minute_second_millis` is really the best fit

PR #737, opened 2024-08-01
[BUG] Deadlock prone code in HttpConnection.cs #689

### LINKED ISSUE #689: [BUG] Deadlock prone code in HttpConnection.cs
### What is the bug?
The HttpConnection class contains the following code in the Request method:
`responseMessage = client.SendAsync(requestMessage, HttpCompletionOption.ResponseHeadersRead).GetAwaiter().GetResult();`
and a little further into the same method:
`responseStream = responseMessage.Content.ReadAsStreamAsync().GetAwaiter().GetResult();`

These two lines of code are prone to deadlocks due to calling GetAwaiter().GetResult(). [This](https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html) link explains why.

A surprisingly similar issue has been discussed to great length in an ElasticSearch.Net [issue](https://github.com/elastic/elasticsearch-net/issues/6342) as well.

It's difficult to reproduce. Essentially requires continuously calling the Request method from multiple threads to provoke a deadlock. It can be quite common under load in a production environment, however.

The code should be free of potential deadlocks. I suggest changing the implementation to use the synchronous "Send" method of the System.Net.HttpClient instead.

DavidPoulsenHGS:
### Description
Fixes deadlock prone code in HttpConnection.cs. Changed synchronous methods in HttpConnection.cs to use Task.Run/Task.Wait to avoid deadlocks in accordance with: https://learn.microsoft.com/en-us/archive/blogs/jpsanders/asp-net-do-not-use-task-result-in-main-context

### Issues Resolved
#689

### DIFF
--- a/src/OpenSearch.Net/Connection/HttpConnection.cs
+++ b/src/OpenSearch.Net/Connection/HttpConnection.cs
@@ -107,8 +107,9 @@ namespace OpenSearch.Net
 
 					if (requestData.ThreadPoolStats)
 						threadPoolStats = ThreadPoolStats.GetStats();
-
-					responseMessage = client.SendAsync(requestMessage, HttpCompletionOption.ResponseHeadersRead).GetAwaiter().GetResult();
+                    var task = Task.Run(() => client.SendAsync(requestMessage, HttpCompletionOption.ResponseHeadersRead));
+                    task.Wait();
+					responseMessage = task.Result;
 					statusCode = (int)responseMessage.StatusCode;
 					d.EndState = statusCode;
 				}
@@ -120,7 +121,9 @@ namespace OpenSearch.Net
 				if (responseMessage.Content != null)
 				{
 					receive = DiagnosticSource.Diagnose(DiagnosticSources.HttpConnection.ReceiveBody, requestData, statusCode);
-					responseStream = responseMessage.Content.ReadAsStreamAsync().GetAwaiter().GetResult();
+                    var task = Task.Run(() => responseMessage.Content.ReadAsStreamAsync());
+                    task.Wait();
+					responseStream = task.Result;
 				}
 			}
 			catch (TaskCanceledException e)


src/OpenSearch.Net/Connection/HttpConnection.cs
```diff
@@ -120,7 +121,9 @@ public virtual TResponse Request<TResponse>(RequestData requestData)
 				if (responseMessage.Content != null)
 				{
 					receive = DiagnosticSource.Diagnose(DiagnosticSources.HttpConnection.ReceiveBody, requestData, statusCode);
-					responseStream = responseMessage.Content.ReadAsStreamAsync().GetAwaiter().GetResult();
+                    var task = Task.Run(() => responseMessage.Content.ReadAsStreamAsync());
```
joefeser, line 124: have you considered just changing it to responseMessage.Content.ReadAsStreamAsync().ConfigureAwait(false)?

src/OpenSearch.Net/Connection/HttpConnection.cs
```diff
@@ -107,8 +107,9 @@ public virtual TResponse Request<TResponse>(RequestData requestData)
 
 					if (requestData.ThreadPoolStats)
 						threadPoolStats = ThreadPoolStats.GetStats();
-
-					responseMessage = client.SendAsync(requestMessage, HttpCompletionOption.ResponseHeadersRead).GetAwaiter().GetResult();
+                    var task = Task.Run(() => client.SendAsync(requestMessage, HttpCompletionOption.ResponseHeadersRead));
```
joefeser, line 110: have you looked at using client.SendAsync(requestMessage, HttpCompletionOption.ResponseHeadersRead).ConfigureAwait(false)?

joefeser (changes requested):

src/OpenSearch.Net/Connection/HttpConnection.cs
Salgat: @joefeser  .ConfigureAwait(false) is not needed inside a Task.Run since there is no synchronization context to capture.

Salgat:
Using .GetAwaiter().GetResult() in a library is considered a massive problem since it does open the door to deadlock. As a last resort, Task.Run(...).Wait() is definitely better than the current .GetAwaiter().GetResult(), but the more appropriate option is to remove the synchronous logic altogether if it's just wrapping an async call in a synchronous wait. At that point, the user of the library can do that themselves using their own judgment on whether it's necessary and with full knowledge of what is happening.

SamuelCox:
Hey all, this is a very old PR, I'd like to attempt to get opensearch-net into a very clean slate so unless someone protests I will delete close this PR within a few days. We can create a new issue or new pr for the underlying issue once we've tackled higher priority stuff.

PR #739, opened 2024-08-05
feat(): centralize the .editorconfig to the root for consistency

joefeser:
### Description
Enables all of the c# projects to have consistent rules and formatting.

### Issues Resolved
Not having to fix tab / space inconsistencies 


.editorconfig
```diff
@@ -3,7 +3,7 @@ root=true
 [*]
 charset = utf-8
 end_of_line = lf
-indent_style = space
+indent_style = tabs
```
dblock, line 6: Isn't this a settled debate where things look different on Windows vs. *nix with tabs? :) 

Can we use 2 spaces?

joefeser: We could do that, but that would require changing the entire project. Every single current file is formatted with tabs, which is why I kept the setting

dblock: Got it. Leaving it to @Xtansia :)

Xtansia: While it's not currently fully consistent as most older files still have tabs, all newer files are indented with spaces, and generated code is opinionatedly formatted with spaces. 

I'd rather go to the effort of unifying things on spaces than to revert this setting, especially given reverting this setting still doesn't fix the inconsistencies.

joefeser: The goal would be to run a dotnet format on everything to get everything consistent. Do we want to also talk about some of the other settings that we want to unify, like no {} on namespaces or any other .editorconfig niceties?

joefeser: I added all of the missing vs.net settings and organized them by the correct section in your .editorconfig file. I made them alph case per section to make it easier to use beyond compare

joefeser: Would you want this in the same PR or a new one @dblock. I will ensure we don't do any style changes as part of the PR. Were you ok with adding the other lines to the .editorconfig?

dblock: I don't have a strong preference, deferring to @Xtansia. We'll take any imrovement!

joefeser: A lot of the noise is around these 3 use cases:
`
//space mismatch on custom attributes on one folder
[DataMember(Name ="path_match")]

[DataMember(Name = "path_match")]

//spacing for object creation
Summaries.Add(new RunSummary {Total = skipReasons.Count, Skipped = skipReasons.Count});

Summaries.Add(new RunSummary { Total = skipReasons.Count, Skipped = skipReasons.Count });   

//class devs for empty classes
public class ValueCountAggregationDescriptor<T>
  : FormattableMetricAggregationDescriptorBase<ValueCountAggregationDescriptor<T>, IValueCountAggregation, T>
    , IValueCountAggregation
  where T : class { }

public class ValueCountAggregationDescriptor<T>
  : FormattableMetricAggregationDescriptorBase<ValueCountAggregationDescriptor<T>, IValueCountAggregation, T>
    , IValueCountAggregation
  where T : class
{ }
`

I am still cleaning up the sample PR of what the outcome of the .editorconfig will be. I need a few hours

Xtansia: From my perspective for ease of review-ability, this will need 3 PRs (not necessarily in this order):
* Change ApiGenerator to invoke `dotnet format` rather than `CSharpier` to ensure styling will be consistent in generated code
* Adjust `.editorconfig`(s) to be consistent and centralized etc.
* Run `dotnet format`

Depending on the style change I'm not particularly fussy (other than keeping `spaces`), I think the change to non-braced namespaces would be fine to include especially given nearly every file is already going to have whitespace/indent changes anyway.

joefeser: PR #747 Created. Adjust ApiGenerator

Xtansia: Given `dotnet format` is very [adverse to line breaking/wrapping/chopping based on line length](https://github.com/dotnet/format/issues/246), there's very little that it does to change formatting outside of some indent and spacing options. The code-style/lint analysis is very useful though. So there might be an argument to use a combination of [CSharpier & dotnet format](https://csharpier.com/docs/IntegratingWithLinters).

I'm going to do a bit of assessment of options and see what's most feasible.

PR #747, opened 2024-08-08
feat(): adjust apigenerator to use dotnet format

joefeser:
### Description
Adjust the generator to use dotnet format

### Issues Resolved
Request from #739, Phase 1


src/ApiGenerator/Program.cs
```diff
@@ -164,9 +169,57 @@ private static async Task<int> Generate(bool download, string branch, bool inclu
...
+            process.ErrorDataReceived += (sender, args) => Console.WriteLine(args.Data);
+
+            process.Start();
+
+            // start our event pumps
+            process.BeginOutputReadLine();
+            process.BeginErrorReadLine();
+
+            process.WaitForExit();
+            Console.WriteLine();
```
Xtansia, line 216: If you add a package reference on `Proc = 0.8.1` which is already used by other projects in the solution, this could be simplified down to something like:
```csharp
            Proc.Exec(new ExecArguments("dotnet", "format", "--verbosity", "normal", "--include", "./src/*/_Generated/")
            {
                WorkingDirectory = GeneratorLocations.SolutionFolder
            })
```

And just define `GeneratorLocations.SolutionFolder` as something like:
```csharp
        public static string SolutionFolder { get; } = $"{Root}../../";
```

joefeser: No problem, thank you for the feedback. I will make the changes this evening

joefeser: PR Comments resolved

joefeser: @Xtansia I updated the pr off the current main branch. If you ignore _generated files, there are less than 20 files to review.

PR #757, opened 2024-08-15
Integration test against recent OpenSearch versions

Xtansia:
### Description
_Describe what this change achieves._

### Issues Resolved
_List any issues this PR will resolve, e.g. Closes [...]._ 


dblock (approved):
So what's the root cause of the failures of integration tests?

Xtansia:
> So what's the root cause of the failures of integration tests?

The main reason for the sudden seemingly unrelated failure of the segment replication test was due to the low disk watermark causing the replica shards to not get assigned to the second node meaning segrep tasks didn't start. There was calls to wait for green cluster health but they weren't actually validating the response. Digging into it caused me to find a handful of other things that were fragile about the multi-node setup and fixed them along the way. Usage of same data folder for both nodes, not setting seed_hosts and setting of initial_cluster_manager wasn't happening before starting the node.

PR #917, opened 2025-04-03
Add support for hybrid query

### LINKED ISSUE #625: Support for HybridQuery
### What are you proposing?
OpenSearch supports Hybrid Query since 2.10 however, using OpenSearch.NET client query DSL we cannot create Hybrid Query today.
[OpenSearch hybrid query](https://opensearch.org/docs/latest/query-dsl/compound/hybrid/)

### LINKED ISSUE #832: [BUG] ModelId should be optional for NeuralQuery
### What is the bug?
If the ModelId is not provided for a Neural query, it is considered conditionless and ignored.  However, this is not a correct implementation since a default ModelId can be configured through a search pipeline.
https://opensearch.org/docs/latest/search-plugins/search-pipelines/neural-query-enricher/

Specify a NeuralQuery without the ModelId

If ModelId is not provided, it should execute the query.

Windows, 10

The fix I tested appears to be simple.  Just update line 54 in NeuralQuery.cs to remove the "string.IsNullOrEmpty(q.ModelId) condition.

### LINKED ISSUE #927: [FEATURE]  Expose max_distance field of knn query
According to the OpenSearch documentation ([https://docs.opensearch.org/docs/latest/query-dsl/specialized/k-nn/](https://docs.opensearch.org/docs/latest/query-dsl/specialized/k-nn/)), k-NN queries have an optional `max_distance` field. This field can replace the `k` field to control which results are returned. `max_distance` is not exposed by the .net client.

Image

Add a `float? MaxDistance` property to the `KnnQuery` class.

You can make `max_distance` work using a Raw Json query, but it kinda defeats the purpose of using the .net client.

While we're at it, might as well add a `float? MinScore` property. Same idea.

Xtansia:
### Description
Add support for hybrid query

### Issues Resolved
Fixes #625 
Fixes #832
Fixes #927

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -71,6 +71,7 @@ jobs:
   integration-opensearch-unreleased:
     name: Integration OpenSearch Unreleased
     runs-on: ubuntu-latest
+    if: false # Temporarily disabled due to k-NN build issues
     strategy:
       fail-fast: false
       matrix:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Changed
 - Changed the namespace client properties on `IOpenSearchClient` to return corresponding interfaces to better enable mocking & unit testing ([#646](https://github.com/opensearch-project/opensearch-net/pull/646))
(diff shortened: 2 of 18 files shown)


Xtansia:
@dblock or @VachaShah Could you please review this PR when you have time?

dblock (approved):

PR #951, opened 2025-08-27
fix diagnostics for async serialize and deserialize

### LINKED ISSUE #950: [BUG] Wrong measurement of SerializeAsync and DeserializeAsync diagnostics
### What is the bug?
SerializeAsync and DeserializeAsync only measure the time it takes to return the `Task`, not the execution time because they don't await in the using block 

Serialize and Deserialze spans should include the execution time even in async mode

sp-jcberleur:
### Description
fix diagnostics for async serialize and deserialize

### Issues Resolved
Closes #950 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Added
 ### Removed
 ### Fixed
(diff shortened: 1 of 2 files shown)


Hailong-am (dismissed):
LGTM — the fix is correct and minimal. The `using` block now properly spans the full async execution.

One thing needed before merge: please add a CHANGELOG entry under `### Fixed` in the `[Unreleased]` section:

```
- Fixed `SerializeAsync` and `DeserializeAsync` diagnostic spans measuring only Task creation time instead of actual execution time ([#950](https://github.com/opensearch-project/opensearch-net/issues/950))
```

Thanks for the contribution!

Hailong-am (approved):

PR #963, opened 2026-05-14
Fix StackOverflowException when serializing KnnVectorProperty via AutoMap visitor.

### LINKED ISSUE #962: [BUG] StackOverflowException when serializing KnnVectorProperty produced via AutoMap visitor
### What is the bug?

`PropertyFormatter.Serialize` is missing a `case IKnnVectorProperty` arm in its dispatch switch ([`src/OpenSearch.Client/Mapping/Types/PropertyFormatter.cs`](https://github.com/opensearch-project/opensearch-net/blob/v1.8.0/src/OpenSearch.Client/Mapping/Types/PropertyFormatter.cs#L140-L225)). When a mapping tree contains a `KnnVectorProperty`, the value falls into the catch-all that asks for the `object` formatter, and Utf8Json's reflection-based fallback walks `PropertyName.Property` → `PropertyInfo` → `Module` → `Assembly` → `Type[]` → `ConstructorInfo[]` forever, terminating the process with `StackOverflowException`.

This is reachable from any application using the documented `AutoMap(IPropertyVisitor)` extension point with a visitor that returns a `KnnVectorProperty` from `Visit(PropertyInfo, OpenSearchPropertyAttributeBase)`. Once returned, `AutoMap` assigns the property's `Name` from the visited `PropertyInfo`, seeding the cycle. The deserialize side already routes `FieldType.KnnVector` → `Deserialize<KnnVectorProperty>`, so the typed formatter exists in the resolver and the serialize side just needs to invoke it.

The descriptor pattern (`.Properties(p => p.KnnVector(kv => ...))`) is unaffected, because it builds the mapping through `KnnVectorPropertyDescriptor` and serializes via a different code path.

Minimal repro: <https://github.com/LeuisKen/opensearch-net-knn-stackoverflow-repro> (40 lines, no live cluster needed).

```csharp
internal sealed class DocWithVector
{
    public float[]? Vec { get; set; }
}

internal sealed class KnnReturningVisitor : NoopPropertyVisitor
{
    public override IProperty? Visit(PropertyInfo propertyInfo, OpenSearchPropertyAttributeBase attribute)
    {
        if (propertyInfo.PropertyType == typeof(float[]))
        {
            return new KnnVectorProperty
            {
                Dimension = 3,
                Method = new KnnMethod
                {
                    Name = "hnsw", Engine = "lucene", SpaceType = "l2",
                    Parameters = new Dictionary<string, object> { ["m"] = 24 },
                },
            };
        }
        return null;
    }
}

var descriptor = new TypeMappingDescriptor<DocWithVector>()
    .AutoMap(new KnnReturningVisitor())
    .Dynamic(false);
var mapping = (ITypeMapping)descriptor;

var client = new OpenSearchClient(new ConnectionSettings(new Uri("http://example.invalid:9200")));
using var ms = new MemoryStream();
client.RequestResponseSerializer.Serialize(mapping, ms); // <-- StackOverflowException
```

Running `dotnet run -c Release` produces:

```
[ok] Mapping tree built; property count: 1
[..] About to serialize. If the bug is present, the process dies with a StackOverflowException now.
Stack overflow.
   at System.Reflection.RuntimeConstructorInfo.get_Name()
   at DynamicClass.Serialize(..., System.Reflection.ConstructorInfo, ...)
   at OpenSearch.Net.Utf8Json.Formatters.CollectionFormatterBase`4[..., ConstructorInfo, ...]
   at DynamicClass.Serialize(..., System.Reflection.TypeInfo, ...)
   at OpenSearch.Net.Utf8Json.Formatters.CollectionFormatterBase`4[..., TypeInfo, ...]
   at DynamicClass.Serialize(..., System.Reflection.Assembly, ...)
   at DynamicClass.Serialize(..., System.Reflection.Module, ...)
   at DynamicClass.Serialize(..., System.Reflection.ConstructorInfo, ...)
   ... (recurses until stack exhaustion)
```

The mapping should serialize to valid JSON (the same JSON the descriptor path produces), e.g.:

```json
{
  "properties": {
    "vec": {
      "type": "knn_vector",
      "dimension": 3,
      "method": {
        "name": "hnsw",
        "engine": "lucene",
        "space_type": "l2",
        "parameters": { "m": 24 }
      }
    }
  }
}
```

### Suggested patch

```diff
                 case IGenericProperty genericProperty:
                     Serialize(ref writer, genericProperty, formatterResolver);
                     break;
+                case IKnnVectorProperty knnVectorProperty:
+                    Serialize(ref writer, knnVectorProperty, formatterResolver);
+                    break;
                 default:
                     var formatter = formatterResolver.GetFormatter<object>();
                     formatter.Serialize(ref writer, value, formatterResolver);
                     break;
```

A defensive secondary fix would be to `throw` from the default arm (rather than silently falling back to reflection), so that any future `IProperty` subtype missed by the switch surfaces a comprehensible error instead of a stack overflow.

Happy to send a PR if maintainers agree on the approach.

| | |
|---|---|
| OpenSearch.Client | 1.8.0 (latest) |
| .NET | net8.0 (also reproduces on net10.0) |
| OS | Windows 11 (CLR-level; OS-independent) |

LeuisKen:
### Description

`PropertyFormatter.Serialize` had no `case IKnnVectorProperty` arm, so the value fell into the reflection-based catch-all and recursed until stack exhaustion. Mirrors the `FieldType.KnnVector` case already present on the `Deserialize` side.

Added a regression test that serializes a visitor-returned `KnnVectorProperty` through `RequestResponseSerializer`; it crashes the xunit host before the fix.

### Issues Resolved

Closes #962.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Fixed
 - Fixed naming of `ClusterManagerTimeout` and `MasterTimeout` properties from `*TimeSpanout` in the low-level client ([#332](https://github.com/opensearch-project/opensearch-net/pull/332))
+- Fixed `StackOverflowException` when serializing a `KnnVectorProperty` returned from a custom `IPropertyVisitor` via `AutoMap` ([#963](https://github.com/opensearch-project/opensearch-net/pull/963))
 
 ### Dependencies
 - Bumps `System.Diagnostics.DiagnosticSource` from 6.0.1 to 8.0.1
--- a/src/OpenSearch.Client/Mapping/Types/PropertyFormatter.cs
+++ b/src/OpenSearch.Client/Mapping/Types/PropertyFormatter.cs
@@ -218,6 +218,9 @@ namespace OpenSearch.Client
 				case IRankFeaturesProperty rankFeaturesProperty:
 					Serialize(ref writer, rankFeaturesProperty, formatterResolver);
 					break;
+				case IKnnVectorProperty knnVectorProperty:
+					Serialize(ref writer, knnVectorProperty, formatterResolver);
+					break;
 				case IGenericProperty genericProperty:
 					Serialize(ref writer, genericProperty, formatterResolver);
 					break;
--- a/tests/Tests/Mapping/Types/Specialized/Knn/KnnVectorVisitorTests.cs
+++ b/tests/Tests/Mapping/Types/Specialized/Knn/KnnVectorVisitorTests.cs
@@ -1,0 +1,70 @@
+/* SPDX-License-Identifier: Apache-2.0
+*
+* The OpenSearch Contributors require contributions made to
+* this file be licensed under the Apache-2.0 license or a
+* compatible open source license.
+*/
+
+using System.Reflection;
+using FluentAssertions;
+using OpenSearch.Client;
+using OpenSearch.Net;
+using OpenSearch.OpenSearch.Xunit.XunitPlumbing;
+using Tests.Core.Client;
+
+namespace Tests.Mapping.Types.Specialized.Knn;
+
+/// <summary>
+/// Regression tests for <see cref="IKnnVectorProperty"/> serialization through the
+/// <see cref="IPropertyVisitor"/> extension point.
+///
+/// Before the corresponding <c>PropertyFormatter</c> fix, the switch in
+/// <c>PropertyFormatter.Serialize</c> had no <c>case IKnnVectorProperty</c> arm, so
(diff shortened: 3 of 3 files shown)


SamuelCox:
Hey, thanks for this. It seems our CI/CD has been failing for some time, so I need to work on fixing that before merging. I'll be in touch once it's done, I only work on this on weekends so may be some time.

Xtansia:
@LeuisKen Could you please rebase onto the latest `main` branch please?

LeuisKen:
> @LeuisKen Could you please rebase onto the latest `main` branch please?

Done.

Xtansia (approved):

Xtansia:
Thanks @LeuisKen!

PR #981, opened 2026-07-03
Add local port forwarding support for AWS SigV4 (#978)

### LINKED ISSUE #978: Add local Port Forwarding Support for AwsSigv4
AWS Open search uses awsSigv4 to authenticate requests to OpenSearch when using IAM authentication.  An OpenSearch cluster can be attached to a private subnet on a VPC, which makes the OpenSearch cluster only accessible from within the VPC.

The AWS recommended solution for outside access, such as during local development, is to SSH to a bastion host, as described under the **Testing VPC Domains** header in: https://docs.aws.amazon.com/opensearch-service/latest/developerguide/vpc.html

A problem arises when using port forwarding to reach OpenSearch over a bastion host due to the SigV4 signing.

Let's say I set up a port forwarding connection through my bastion host on my local machine from local port 9200 to 443 on the AWS VPC.  I add a /etc/hosts entry to point **search‑mycompany‑prod‑uswest2.es.amazonaws.com** to 127.0.0.1, so the SSL SNI will be correct.

Now I can connect to the OpenSearch cluster using **search‑mycompany‑prod‑uswest2.es.amazonaws.com:9200** on my local machine.

In this setup, AWS Sigv4 signing fails with an error like the following (shortened for brevity):

```
    The Canonical String for this request should have been\n'
    GET accept:application/json
    host:search‑mycompany‑prod‑uswest2.es.amazonaws.com
    opensearch-client-meta:opensearch=1.8.0,a=1,net=10.0.9,so=10.0.9
```

The issue is that the request was signed with **search‑mycompany‑prod‑uswest2.es.amazonaws.com:9200**, but the error from AWS shows that they are not using the port number when building the canonical string to verify the signature on the request.

I found a similar issue on an unrelated project:
https://github.com/opensearch-project/opensearch-go/issues/370

This scenario can work if the port number is assigned to the HTTP Uri after the request is signed, but before it is sent.

I was able to do it by adding the port number to the outgoing HTTP request here.

https://github.com/opensearch-project/opensearch-net/blob/3c5d1e0da8ec5eb3ca19f98d08199b70d660e60f/src/OpenSearch.Net.Auth.AwsSigV4/AwsSigV4HttpClientHandler.cs#L38

I had to do a local implementation of AwsSigV4HttpClientHandler, since it is internal.

```csharp
   // This is what I did:
    protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
    {
        var credentials = await _credentials.GetCredentialsAsync().ConfigureAwait(false);
        // search‑mycompany‑prod‑uswest2.es.amazonaws.com is the original URI for the request, used here for signing
        await AwsSigV4Util.SignRequest(request, credentials, _region, _dateTimeProvider.Now(), _service).ConfigureAwait(false);

        // Now we add the port number, so the request will go out on port 9200
        if (_tunnelPortNumber != null) { 
        
            var builder = new UriBuilder(request.RequestUri!) { Port = (int)_tunnelPortNumber};
            request.RequestUri = builder.Uri;
        }
        
        //send it to port 9200, which will forward to 443 on the VPC
        return await base.SendAsync(request, cancellationToken).ConfigureAwait(false);
    }

````

My request is that a tunnel port number configuration option be added to allowed a setup like this in https://github.com/opensearch-project/opensearch-net/blob/main/src/OpenSearch.Net.Auth.AwsSigV4/AwsSigV4HttpConnection.cs

I realize I could use Basic Auth to authenticate with OpenSearch, but I would prefer to use IAM permissioning and be able to do it locally for local .Net app development.

I think it would also work right now if I used 443 for my local forwarding port, but I had trouble with that in my corporate environment.

I think curl also has solutioning for this situation with the --aws-sigv4 argument and specifying the Host header?

### PR DESCRIPTION
**Pull request by YiyuanMiao**
### Description
  Adds an optional `TunnelPort` property to `AwsSigV4HttpConnection`. When set, the request is signed against its canonical host (without the port) so the
  signature matches what AWS computes, then the port is applied to the request URI *after* signing but *before* dispatch — so the request reaches a local
  tunnel (e.g. SSH local port forwarding through a bastion host) while the signature stays valid.

  ### Issues Resolved
  Closes #978

  ### Testing
  Added `TunnelPortTests` verifying that with a tunnel port set, the dispatched request URI carries the port while the SigV4 signature is identical to the
  no-port case; and that behaviour is unchanged when no tunnel port is set.

  ### Check List
  - [x] New functionality includes testing.
  - [x] New functionality has been documented (CHANGELOG).

### CODE REVIEW

File: src/OpenSearch.Net.Auth.AwsSigV4/AwsSigV4HttpConnection.cs [hunk 1]
```diff
@@ -20,6 +20,15 @@ public class AwsSigV4HttpConnection : HttpConnection
...
 		public const string OpenSearchServerlessService = "aoss";
 
+		/// <summary>
+		/// An optional port to apply to each request's URI <em>after</em> it has been signed. This is useful when
+		/// tunnelling to a cluster on a private subnet (e.g. via SSH local port forwarding through a bastion host):
+		/// the signature is computed against the canonical host without the local tunnel port, matching what AWS
+		/// verifies, while the request is still dispatched to the local tunnel port. Leave as <c>null</c> (the default)
+		/// to disable this behaviour.
+		/// </summary>
+		public int? TunnelPort { get; set; }
```
SamuelCox: Is there any reason for this to be mutable?
Seems better for it to be a private readonly in following with the conventions of the rest of the file. There's no reasonable situation where you'd change it after it's set, I'd assume?

Xtansia:
@YiyuanMiao 
Focusing solely on a "tunnel port" is too narrow and doesn't solve the problem in general, especially when mutating the request URI within the connection implementation. There are also cases where the tunnel host may be different too (i.e. without modifying `/etc/host` and just connecting via `http://localhost:9200` tunnelled to AWS). 

Something that may be worth experimenting with, as in [opensearch-rs](https://github.com/opensearch-project/opensearch-rs/blob/2.4/opensearch/tests/aws_auth.rs#L58), is to configure the true request URI on the client (i.e., with the port or a different host), and then override the host header, which is handled by the signer.
i.e. tweaking https://github.com/opensearch-project/opensearch-net/blob/main/src/OpenSearch.Net.Auth.AwsSigV4/CanonicalRequest.cs#L69 to try pulling a `Host` header out of `request.Headers`, otherwise falling back to `request.RequestUri.Authority`. 
Otherwise, rather than passing a tunnel port to the connection, just allow direct overriding of the signing host value.

YiyuanMiao:
Thanks both for the feedback — you're right that the tunnel-port approach was too narrow.
@Xtansia I've reworked this along the lines you suggested. Instead of adding a dedicated TunnelPort property and mutating the request URI after signing, CanonicalRequest now signs against the caller-supplied Host header when one is present, falling back to request.RequestUri.Authority otherwise (the change is right at the line you pointed to). This handles both the host and the port mismatch in one place — since the port is part of the authority anyway — and covers the "connect directly to http://localhost:9200" case without touching /etc/hosts. Users just point the connection at the tunnel and override the host via the existing GlobalHeaders API, eg:
var settings = new ConnectionSettings(new Uri("http://localhost:9200"), connection)
      .GlobalHeaders(new NameValueCollection { { "Host", "xxx.es.amazonaws.com" } });

I've replaced TunnelPortTests with SigningHostOverrideTests.
@SamuelCox this also resolves your point about the property's mutability — TunnelPort has been removed entirely, so no longer applies.

Xtansia:
The change looks good @YiyuanMiao, were you able to validate this approach works for your SSH tunnel setup?

YiyuanMiao:
> The change looks good @YiyuanMiao, were you able to validate this approach works for your SSH tunnel setup?

Sorry for replying late. I had some other opensearch works to do last week.
I validated end-to-end against a real managed OpenSearch domain (`es`, us-west-2). I emulated the port-forwarding tunnel with a local forwarder (`socat TCP-LISTEN:9200 →
  <domain>:443`), pointed the client at `https://localhost:9200`, and overrode the `Host` header to the real endpoint via `GlobalHeaders`:

  - **With** the Host override → `Ping` returns **200** (request dispatched through the tunnel to localhost, signature computed for the real endpoint).
Screenshot 2026-07-16 at 11 41 32

  - **Without** it (signing against `localhost:920
Screenshot 2026-07-16 at 11 42 22
0`) → **403 SignatureDoesNotMatch**.
  
  This confirms the request reaches the tunnel while the signature stays valid for the real endpoint — exactly the scenario from #978.

   ## How a user would use this

  No new API — the signing host is just supplied through the existing `GlobalHeaders`. Given an SSH local port forward to a cluster on a private subnet:

  ```bash
  ssh -N -L 9200:my-domain.us-west-2.es.amazonaws.com:443 user@bastion-host
  ```

  Then, in your **application code** where you set up the client (this is the user's own code, not part of the library):
  
  ```csharp
  var connection = new AwsSigV4HttpConnection();
  var settings = new ConnectionSettings(
          new Uri("https://localhost:9200"),                     // dispatch to the tunnel
          connection)
      .GlobalHeaders(new NameValueCollection {
          { "Host", "my-domain.us-west-2.es.amazonaws.com" }     // sign for the real endpoint
      });
  var client = new OpenSearchClient(settings);
  ```

  The request goes to `localhost:9200` (the tunnel) while the SigV4 signature is computed against `my-domain.us-west-2.es.amazonaws.com`, so AWS accepts it. When no `Host` header is
  set, behaviour is unchanged (signs against the request URI's authority).

YiyuanMiao:
I've just rebased/merged main and resolved the conflicts, so it's mergeable again and CI is green. Would any of the maintainers have a chance to take a look when you get a moment? Happy to make any changes. Thanks!

Xtansia (approved):

YiyuanMiao:
This CI failure is a pre-existing flaky integration test, **unrelated to this PR — re-running should clear it**
@Xtansia @SamuelCox @VachaShah @Yury-Fridlyand @MaxKsyunz

  The integrate job failed on a single test out of 1463:

  Failed search.aggregation 20_terms.yml
    - "Mixing longs, unsigned long and doubles"
    - Assert Match hits.total  expected: 6.0  actual: 28.0

  This is not a regression from this PR. The aggregation itself is fully correct — the returned buckets sum to exactly the 6 documents the section indexes:

  "buckets":[
    {"key":10.0,"doc_count":2},
    {"key":100.0,"doc_count":2},
    {"key":14.6,"doc_count":1},
    {"key":1.0223372036854776E19,"doc_count":1}   // 2+2+1+1 = 6
  ]

  **The only failing assertion is hits.total, which returned 28 instead of 6.** That section issues **a cluster-wide _search with no index specified, so hits.total counts documents across every index in the cluster**. 
Under the readonly, multinode, writable integrate run, other suites are writing to the same cluster concurrently, so the extra 22 docs are leakage from parallel tests — a timing-dependent race, not a client-side defect. This is a classic flaky failure: same code, different outcome depending on when the query happens to run relative to other suites.

  Notably, this branch makes no changes to tests/Tests.YamlRunner/ or the skip list, and the same file already skips two sibling sections for related unsigned-long reasons (SkipList.fs).

  Suggested action: **please re-run the failed job — it should go green.** If it proves persistently flaky, the consistent fix would be to add "Mixing longs, unsigned long and doubles" to the existing skip entry for 20_terms.yml (it's a cluster-wide _search whose hits.total isn't reliable under concurrent integrate), but that's a test-infra decision for the maintainers rather than something this PR should carry.

PR #979, opened 2026-06-27
[DRAFT] System.Text.Json serialization foundation (#388)

xluo-aws:
> **Draft / RFC.** Foundation + de-risking spike for migrating serialization off the vendored, unmaintained Utf8Json to System.Text.Json. Opening early for design feedback before the large converter-generation effort. Not for merge as-is.

Relates to #388 (also #370 trimming, #424 AOT, #318).

### Why
`OpenSearch.Net` compiles a vendored copy of the abandoned **Utf8Json** (57 files, ~20.5k LOC) as the default internal serializer; the high-level client is coupled to it via ~151 `IJsonFormatter` implementations and ~135 files threading `IJsonFormatterResolver`. This blocks enterprise security reviews and trimming/AOT. A source-serializer-only package does **not** solve it — the internal engine must be replaced.

### What this PR establishes (foundation, additive & non-breaking — Utf8Json remains the default)
1. **`SystemTextJsonSerializer`** in `OpenSearch.Net` implementing the existing `IOpenSearchSerializer` (sync/async, generic & non-generic, formatting hint; netstandard2.0-compatible; STJ added for the netstandard targets only).
2. **Low-level wiring** — usable today via `ConnectionConfiguration(pool, connection, new SystemTextJsonSerializer())`; request path runs on STJ.
3. **High-level seam** — `ConnectionSettingsBase.CreateDefaultRequestResponseSerializer()` (protected virtual, defaults to Utf8Json) so STJ can become the high-level internal serializer by override.
4. **Wire-format parity** — tests capture ground-truth JSON from the real Utf8Json client (`match_all`, `term`) and assert an STJ converter reproduces it byte-for-byte.

### De-risking outcome
The four foundational unknowns (incl. the converter **state-threading** that stalled the 2023 attempt, and **wire-format reproducibility**) are now validated with running code.

### Testing
13 xUnit tests under `poc/OpenSearch.Net.Stj.Tests` (a verification harness; real tests would live in the production suite). `OpenSearch.Net` builds across all 5 TFMs; `OpenSearch.Client` unaffected. Note: integration tests against live OpenSearch were not run in this environment.

### What remains (the volume, not the risk)
- Generate/port the ~151 converters so an STJ-backed `OpenSearchClient` reproduces the full wire contract.
- Validate against the full ~3,200-test serialization oracle + live-cluster integration.
- AOT/source-gen path; opt-in legacy Utf8Json for one major; remove the engine.

[...]

Issue #388, opened 2023-10-09
[FEATURE] Add support for System.Text.Json serialization


Scrambles56:
### Is your feature request related to a problem?
As a developer, I'm trying to reduce the amount of dependencies I'm taking on in my projects, and choose to use Microsofts default JSON serializer, System.Text.Json.

### What solution would you like?
An implementation of the `IOpenSearchSerializer` that uses System.Text.Json, that provides equivalent behaviour to the existing Newtonsoft implementation. 

### What alternatives have you considered?
At the moment, the fallback is very isolated usage of the Newtonsoft package.

### Do you have any additional context?
N/A

Xtansia:
I've expanded the scope of this issue to instead be migrating away from `Utf8Json` to `System.Text.Json` as the primary serialization library of the clients.

Djcarrillo6:
Hi @Xtansia, I'd like to request assignment on this issue.

Xtansia:
@Djcarrillo6 You're more than welcome to have a go at this issue

goldenc:
any update on this? we would like to move away from Newtonsoft.Json but its a bit tricky without a proper System.Text.Json open search implementation

Xtansia:
> any update on this? we would like to move away from Newtonsoft.Json but its a bit tricky without a proper System.Text.Json open search implementation

@goldenc There's been no updates yet to the best of my knowledge, I'm unsure if @Djcarrillo6 made any progress. So would be great if anyone was willing to make contributions towards this goal.

Djcarrillo6:
Yeah @Xtansia I've been running into road block after road block. If anyone else would like to also work on this, please feel free.

Hailong-am:
## Utf8Json → System.Text.Json Migration Plan

**Goal:** Replace the vendored, unmaintained Utf8Json library with Microsoft's System.Text.Json (STJ), with a clear path to full AOT / Native AOT support.

---

### Phase 0 — STJ Opt-In (2.x)

Introduce STJ as a fully supported alternative without changing default behavior.

- Utf8Json remains the default serializer — zero breaking change
- `OSC_USE_STJ=true` environment variable switches to STJ
- `settings.UseSystemTextJson()` programmatic API does the same
- CI runs the full test suite under both engines to catch regressions on either path

**User impact:** Existing code is unaffected. Users can opt in to STJ, validate behavior, and report issues before the 3.0 default flip.

---

[...]

PR #982, opened 2026-07-04
Migrate high-level client serialization to System.Text.Json (#388)

xluo-aws:
## Summary

Migrates the **OpenSearch.Client** high-level serializer from the vendored, unmaintained Utf8Json to **`System.Text.Json`**, and flips the client default over to it. Relates to #388.

This is the cutover on top of the incremental converter work: the request/response serializer and a distinct camel-casing source serializer are now both `System.Text.Json`-based.

## Reviewer guide

This PR is stacked on top of the foundation PR #979 and its diff against `main` therefore **includes #979's commits**. To review only the incremental work, please **review #979 first**, then use the incremental-only diff:

- **Incremental diff (this PR minus the foundation)**: https://github.com/xluo-aws/opensearch-net/compare/feature/stj-foundation...feature/stj-request-remainders (93 commits)

Suggested reading order:
1. `dev-docs/system-text-json-migration.md` — design rationale, one section per slice (start here).
2. `src/OpenSearch.Client/CommonAbstractions/SerializationBehavior/SystemTextJsonOptionsFactory.cs` — the central registry mapping every converter into the two options instances (the "map" of the PR).
3. `src/OpenSearch.Net/Serialization/DataContractResolver.cs` — the core mechanism (honoring `[DataMember]` et al.).
4. Cutover wiring in `src/OpenSearch.Client/CommonAbstractions/ConnectionSettings/ConnectionSettingsBase.cs`.
5. `src/OpenSearch.Client/Aggregations/AggregateConverter.cs` — the heuristic response reader (the most intricate piece; worth the closest look).

The many small per-type converters preserve wire output **byte-for-byte** vs. Utf8Json (validated per slice + full CI), so they can be skimmed for structure rather than re-derived. Reviewing commit-by-commit is easier than the unified diff — commits are scoped one slice/fix each.

## What's included

- **Serializer infrastructure**: interface-aware `DataContractResolver` (honors `[DataMember]`/`[IgnoreDataMember]`/`ShouldSerialize`/non-public setters/ctor-less types/explicit interface impls/`[ReadAs]`), `StringEnumConverterFactory`, `ObjectConverter` for dynamic payloads, Utf8Json-compatible number formatting, settings-threaded `Field`/`PropertyName`/`Id`/`Indices`/`Routing` converters, and a central options factory.
- **Request DSL**: analysis, scripts, the full query DSL, request-side aggregations, and mapping properties.
- **Value types & mechanisms**: per-property (exact + open-generic) converter overrides, verbatim dictionaries, single-or-enumerable, ingest processors, `dynamic_templates`, index settings (flatten/unflatten), cluster reroute, and the cross-cutting value types.
- **Response readers**: aggregation responses, leaf value readers, dictionary/dynamic response family, bulk response items, suggest dictionary, `GetRepository`, `LazyDocument`/`FieldValues`, and the typed document `_source` reader.
- **Cutover**: default request/response serializer flipped to `System.Text.Json`; a distinct source serializer with document field-name inference (camel-casing + property mappings/attributes); `multi_search`/`multi_get` response builders; and STJ writers for every custom request body -- `_bulk` and `_msearch`/`_msearch/template` (NDJSON, via a self-serialization hook since NDJSON cannot use STJ's single-root writer), plus `index`/`create`, `mget`, `put_settings`, and `create_repository`.

## Testing

- Each slice was validated **byte-for-byte** against the Utf8Json output (isolated harnesses; not shipped).
- **All CI workflows have passed on this branch** (code-complete at commit `67b6365f4`): Compile, Code Generation, unit **Tests**, **Integration tests across all supported OpenSearch versions** (1.1.0 -> 3.6.0), **YAML REST tests**, License headers, and Changelog.
- **Integration flakiness note**: the integration matrix runs 14 clusters in parallel and can occasionally go red for environmental reasons — transient cluster `503`s under load, which surface as empty-hits assertion failures and clear on re-run (not code regressions; e.g. a docs-only commit can show a red integration check over identical, previously-green code). The one seed-dependent test (`PartitionTermsAggregationUsageTests`) has been made deterministic.
- The integration run surfaced a set of real-cluster response-deserialization gaps the isolated harnesses could not (see below); these have been fixed and the suite passes end-to-end.
- Design notes and per-slice details are in `dev-docs/system-text-json-migration.md` (sections 54-56 cover the post-cutover unit and integration fixes).

## Post-cutover fixes surfaced by CI

- **Unit suite**: a long tail of behavioural gaps vs. Utf8Json (top_hits, filters model, value tuples, GeoLocation WKT, custom `IProperty`, long bucket-key precision, completion-suggester member ordering, decimal formatting, etc.).
- **Integration suite** (real responses): dictionary key converters (`IndexName`/`Id`/`TaskId`), `IsADictionary`/`ResolvableDictionaryProxy` construction, non-`IConvertible` dictionary keys, completion of the polymorphic aggregate reader (scripted_metric, pipeline `keys`, geo_bounds/geo_centroid incl. no-results, matrix_stats, composite, significant_terms/text, multi-terms, variable-width, anonymous filters), `Properties` key resolution, `ISuggestContext`/`IFieldMapping` converters, and delegation of inference identity types in the JSON.NET source serializer.

## Not yet done / follow-ups

- Vendored Utf8Json is **not yet removed** (kept until this lands; removal is a follow-up).
- Deferred: epoch-millis per-property date converter.

Draft for review.


xluo-aws:
@SamuelCox @Xtansia @joshuali925 @saratvemulapalli @MaxKsyunz @VachaShah @Yury-Fridlyand @anirudha , do we want a transitional "fall back to Utf8Json" toggle for the STJ cutover? The major cost is it will double the CI test effort since we need to make sure both serializer works and delay cleanup/removal of Utf8Json. The alternative(no toggle) rely on revert back to the last pre-STJ release.

SamuelCox:
> @SamuelCox @Xtansia @joshuali925 @saratvemulapalli @MaxKsyunz @VachaShah @Yury-Fridlyand @anirudha , do we want a transitional "fall back to Utf8Json" toggle for the STJ cutover? The major cost is it will double the CI test effort since we need to make sure both serializer works and delay cleanup/removal of Utf8Json. The alternative(no toggle) rely on revert back to the last pre-STJ release.

When you say double the CI test effort, do you mean double the amount of time the tests take to execute because it will duplicate every test?

Or is there some other testing effort that has a human cost or etc? From my POV, a feature toggle is much safer, and easy, but I don't pay the github actions bill.

xluo-aws:
@SamuelCox , yes, it's most about CI test effort because it need to test both serializer. 

> 
> When you say double the CI test effort, do you mean double the amount of time the tests take to execute because it will duplicate every test?
> 
> Or is there some other testing effort that has a human cost or etc? From my POV, a feature toggle is much safer, and easy, but I don't pay the github actions bill.

SamuelCox:
> @SamuelCox , yes, it's most about CI test effort because it need to test both serializer.
> 
> > When you say double the CI test effort, do you mean double the amount of time the tests take to execute because it will duplicate every test?
> > Or is there some other testing effort that has a human cost or etc? From my POV, a feature toggle is much safer, and easy, but I don't pay the github actions bill.

Yeah as I say, personally I'm completely fine with doubling the CI test effort, but equally I'm not the one that pays the github action bills for the opensearch project. If its purely a technical decision, I'm happy with a feature flag.

SamuelCox:
Let me know if this is ready for review please @xluo-aws , I've got a little bit confuse between the 3 system.text.json PRs I've seen which one is the correct/first one to review

xluo-aws:
@SamuelCox , we have three team members including me trying different solutions now. Mine is to keep Utf8Json for a while then deprecate it so it includes logic to backward compatible between those two serializers, another approach is to fully remove Utf8Json so the code is more clean. Since it's a huge change, we want to make sure all tests pass first so we submit multiple PRs. We'll get aligned first then ask for review. But you are always welcome to comment on the different approach.

xluo-aws:
I have another [PR](https://github.com/xluo-aws/opensearch-net/pull/1/changes) to add a toggle to switch between those two serializers, default will be STJ. But it depends on our final strategy whether to fully remove Utf8Json or keep a while in case STJ has some issues. Although switch back to previous release is an option as well. 

> Yeah as I say, personally I'm completely fine with doubling the CI test effort, but equally I'm not the one that pays the github action bills for the opensearch project. If its purely a technical decision, I'm happy with a feature flag.

src/OpenSearch.Net/Serialization/DataContractResolver.cs
```diff
@@ -0,0 +1,599 @@
...
+				// converter for F, if any (#388). Used for primitives the server may send as strings and
+				// for generic wrappers such as single-or-enumerable members.
+				// A member-level [JsonFormatter] takes precedence; otherwise honor a type-level
+				// [JsonFormatter] on the member's declared type (e.g. the verbatim-keys dictionaries
+				// such as IAnalyzers carry the attribute on the interface, not the member).
+				var formatter = GetAttribute<Utf8Json.JsonFormatterAttribute>(member, interfaceProps)
+					?? member.PropertyType.GetCustomAttribute<Utf8Json.JsonFormatterAttribute>(false);
+				if (formatter != null && TryGetPropertyConverter(formatter.FormatterType, member.PropertyType, out var propertyConverter))
+					property.CustomConverter = propertyConverter;
+			}
```
Hailong-am, line 277: This will make STJ depends on Utf8Json, once we remove utf8json, the logic will break.

xluo-aws: > This will make STJ depends on Utf8Json, once we remove utf8json, the logic will break.

The migration keeps the existing [JsonFormatter]/[InterfaceDataContract] attributes as shared metadata and bridges them to STJ converters, avoiding a ~600-file re-annotation in this PR. STJ depends on the Utf8Json attribute/formatter types as metadata keys, but never on Utf8Json logic (those formatters are never instantiated). Removing Utf8Json is a compile-time re-annotation that deletes these type references together with the bridge — no behavioral risk, and any dangling reference is a hard compile error, not a silent break. Will have phase 2 to remove all Utf8json related code once phase 1(this PR) proves to be stable.

Hailong-am:
> I'm happy with a feature flag.
With a feature flag sounds more safer. How we are going to introduce the flag? The possible ideas in my mind would be 
1. A env variable
2. Per client setting like `new ConnectionSettings().useSTJ()` to switch to STJ serializer

Also with a feature flag, we can't remove Utf8Json completely in 2.x

xluo-aws:
> > I'm happy with a feature flag.
> > With a feature flag sounds more safer. How we are going to introduce the flag? The possible ideas in my mind would be
> 
> 1. A env variable
> 2. Per client setting like `new ConnectionSettings().useSTJ()` to switch to STJ serializer
> 
> Also with a feature flag, we can't remove Utf8Json completely in 2.x

Yes, we can remove Utf8Json completely in 3.0. Env variable may be better because it doesn't require customer to update the code. @SamuelCox , what's your preference?

SamuelCox:
> > > I'm happy with a feature flag.
> > > With a feature flag sounds more safer. How we are going to introduce the flag? The possible ideas in my mind would be
> > 
> > 
> > 
> > 1. A env variable
> > 2. Per client setting like `new ConnectionSettings().useSTJ()` to switch to STJ serializer
> > 
> > Also with a feature flag, we can't remove Utf8Json completely in 2.x
> 
> Yes, we can remove Utf8Json completely in 3.0. Env variable may be better because it doesn't require customer to update the code. @SamuelCox , what's your preference?

I think an env var would be preferable, personally

Xtansia:
Changing the default serializer to STJ in 2.x now that it has been released is somewhat risky in that there could be uncaught breaking changes that you'd be making within a feature release. 
Aside from that I think configuration should still be available programmatically, even if it falls back to reading an env var when unset, as it's easier to discover and be certain of.

xluo-aws:
> Changing the default serializer to STJ in 2.x now that it has been released is somewhat risky in that there could be uncaught breaking changes that you'd be making within a feature release. Aside from that I think configuration should still be available programmatically, even if it falls back to reading an env var when unset, as it's easier to discover and be certain of.

make sense, leave default serializer as Utf8Json won't cause any unexpected surprise to customer after they upgrade to 2.1. The original idea to set the default to STJ is everyone can start using this better maintained Serializer by default after upgrading to latest release 2.1.  @SamuelCox , @joshuali925 ,  any suggestion on which Serializer to be used as the default in 2.1? 

@Xtansia , is there a client configuration file we can add this feature toggle?

src/OpenSearch.Net/Serialization/DataContractResolver.cs
Hailong-am: Make sense, we will need to have another custom attribute to replace it or we replace it directly to `JsonConverter`

SamuelCox:
> > Changing the default serializer to STJ in 2.x now that it has been released is somewhat risky in that there could be uncaught breaking changes that you'd be making within a feature release. Aside from that I think configuration should still be available programmatically, even if it falls back to reading an env var when unset, as it's easier to discover and be certain of.
> 
> make sense, leave default serializer as Utf8Json won't cause any unexpected surprise to customer after they upgrade to 2.1. The original idea to set the default to STJ is everyone can start using this better maintained Serializer by default after upgrading to latest release 2.1. @SamuelCox , @joshuali925 , any suggestion on which Serializer to be used as the default in 2.1?
> 
> @Xtansia , is there a client configuration file we can add this feature toggle?

If its a new release I'm happy for it to be STJ, although, makes me wonder whether it should be a major version increase since its such a big change, I know that it technically shouldn't be a breaking change but it could be

xluo-aws:
> If its a new release I'm happy for it to be STJ, although, makes me wonder whether it should be a major version increase since its such a big change, I know that it technically shouldn't be a breaking change but it could be

@peterzhuamazon , Peter, do you know if we have any restriction on releasing 2 major versions in 1-2 months? I guess not, but want to double check.

Hailong-am:
@SamuelCox @Xtansia @xluo-aws I just post a plan for the whole migration to STJ and have AOT supported in issue https://github.com/opensearch-project/opensearch-net/issues/388#issuecomment-5055962921. Let's discuss the plan in that issue, and this PR will used for phrase 0 implementation.

xluo-aws:
Closed this PR, a better solution is implemented in https://github.com/opensearch-project/opensearch-net/pull/1002

Issue #968, opened 2026-05-21
[BUG] AwsSigV4HttpConnection produces 401 against AOSS when running on AWSSDK.Core 4.x with EKS Pod Identity


ethanejones:
### What is the bug?
`OpenSearch.Net.Auth.AwsSigV4` 1.8.0 produces requests that AWS OpenSearch Serverless (AOSS) rejects with HTTP 401 "Could not authenticate with the specified node" when the host application is running on `AWSSDK.Core` 4.x and obtaining credentials via EKS Pod Identity.

The same role, signing the same request against the same endpoint using `awscurl` (which uses a modern boto3 / AWSSDK signer) succeeds with HTTP 200. So the credentials, IAM role, AOSS data access policy, and region are all confirmed correct. The 401 only manifests through the `AwsSigV4HttpConnection` signing path.

--- 

### How can one reproduce the bug?
Project setup:
  - `net8.0` application
  - `OpenSearch.Net.Auth.AwsSigV4` 1.8.0
  - `AWSSDK.Core` 4.0.3.8 (any 4.0.x)
  - `AWSSDK.OpenSearchServerless` 4.0.5.5
  - Hosted in an EKS pod that obtains credentials via EKS Pod Identity (not IRSA)

Construct the client roughly as follows:
```csharp
var region = RegionEndpoint.GetBySystemName("us-east-1");
var pool = new SingleNodeConnectionPool(new Uri(collectionEndpoint));
var connection = new AwsSigV4HttpConnection(
    region,
    service: AwsSigV4HttpConnection.OpenSearchServerlessService);

var settings = new ConnectionSettings(pool, connection);
var client = new OpenSearchClient(settings);

await client.CountAsync<MyDoc>(c => c.Index("documents-<id>"));
```

Result - every request fails with:
>OpenSearch.Net.OpenSearchClientException: Could not authenticate with the specified node.
Try verifying your credentials or check your Shield configuration..
Call: Status code 401 from: POST `/documents-<id>/_count`
 ---> OpenSearch.Net.PipelineException: Could not authenticate with the specified node.

Control test (same pod, same role) - signing manually with a modern AWS signer succeeds:
```bash
ENDPOINT=$(aws opensearchserverless batch-get-collection \
  --names <collection-name> --region us-east-1 \
  --query 'collectionDetails[0].collectionEndpoint' --output text)

awscurl --service aoss --region us-east-1 \
  -X POST "$ENDPOINT/documents-<id>/_count" \
[...]

PR #987, opened 2026-07-09
Decouple AwsSigV4 signing from the AWS SDK internal signer

xluo-aws:
### Description

`AwsSigV4HttpConnection` computed the SigV4 signature via the AWS SDK **internal** API `Amazon.Runtime.Internal.Auth.AWS4Signer`. Internal SDK APIs are not part of the public contract and carry no cross-version compatibility guarantee. This change reimplements the (small, stable, publicly documented) SigV4 algorithm directly on top of `System.Security.Cryptography` (`SHA256`/`HMACSHA256`) plus the public `AWSSDKUtils` path/URL helpers, removing all reliance on `Amazon.Runtime.Internal.*`.

This is **hardening**: it makes signing robust to `AWSSDK.Core` major-version changes and eliminates a dependency on undocumented internals, which have a far weaker backward-compat guarantee than .NET's own crypto primitives.

### Relationship to #968

This PR is **related to** #968 but, after investigation, is **not** the root-cause fix for that report:

- The reported symptom is an HTTP 401 from Amazon OpenSearch Serverless under **EKS Pod Identity**.
- Reproduction (matching `AWSSDK.Core 4.0.3.8` against the released 1.8.0 binary, static and session-token credentials) shows the pre-fix signing produced **byte-identical, spec-correct signatures** — so the 401 is not a signing bug. Thanks to @Hailong-am for independently reaching the same conclusion.
- The likely real cause is **credential resolution**: 1.8.0 discovered default credentials via the legacy `FallbackCredentialsFactory`, which does not include the EKS Pod Identity provider. The 4.x migration (#931) already switched this to `DefaultAWSCredentialsIdentityResolver`, which does — so the actual Pod Identity fix rides along with the 4.x bump on `main`, independent of this PR. (This is analysis/context; nothing in this diff substantiates it.)

This change is still worth shipping on its own merits (robustness + no internal-API coupling), and it does not regress signing for any current caller.

### Changes

- `src/OpenSearch.Net.Auth.AwsSigV4/AwsSigV4Util.cs`: new `ComputeAuthorizationHeader` implementing the SigV4 string-to-sign, signing-key derivation, and `Authorization` header directly (public crypto only).
- `src/OpenSearch.Net.Auth.AwsSigV4/CanonicalRequest.cs`: replace internal `AWS4Signer.ComputeHash`/`FormatDateTime` with `SHA256` and invariant-culture date formatting; keep the public `AWSSDKUtils` helpers. No internal `Amazon.Runtime.Internal.*` usage remains in the package.
- `src/OpenSearch.Net.Auth.AwsSigV4/OpenSearch.Net.Auth.AwsSigV4.csproj`: add `InternalsVisibleTo` for `Tests.Auth.AwsSigV4` (strong-name key) so the reference-vector test can call the internal helper. (Modifies the shipped project file.)
- `tests/Tests.Auth.AwsSigV4/AwsSigV4HttpConnectionTests.cs`: reference-vector, non-UTC, and session-token tests.
- `tests/Tests.Auth.AwsSigV4/CanonicalRequestTests.cs`: multi-value header canonicalization test.
- `CHANGELOG.md`: hardening entry under **Changed**.

### On signature stability and UTC normalization

Signatures are **byte-for-byte identical to the previous implementation for all `DateTimeKind`s** — including a custom `IDateTimeProvider` that returns a `Local`/`Unspecified` `DateTime`. The new code calls `signingTime.ToUniversalTime()` explicitly; this is **not** a behavior change, because the previous internal signer already normalized to UTC (`AWS4Signer.FormatDateTime` for `x-amz-date`, and `AWS4Signer.ComputeSignature` for the credential-scope datestamp). I verified this by reflecting into `AWSSDK.Core` 4.x: for a `Local` signing time the old `FormatDateTime`/`ComputeSignature` produced exactly the UTC-normalized output, matching an independent spec computation. The `SignsRequestCorrectlyWithNonUtcSigningTime` test pins this equivalence.

### Testing

Validated by:

- Existing `CanonicalRequestTests` and `AwsSigV4HttpConnectionTests` (unchanged expected values).
- `ComputeAuthorizationHeaderMatchesPublishedAwsReferenceVector` — pins signing to AWS's **published** SigV4 reference example (IAM `ListUsers`), independent of any `AWSSDK.Core` version.
- Edge-case coverage added during review: **session token** (temporary credentials — the EKS Pod Identity credential shape), **non-UTC signing time** (UTC-normalization equivalence), and **multi-value headers**. Empty-body and special-character-path cases were already covered.

Note: signature comparison is not applicable — this code *produces* signatures and has no verification/comparison path, so there is no timing side-channel.

### Note on AWSSDK.Core version floor

Loosening the `AWSSDK.Core` floor below 4.x is **not** feasible: `AwsSigV4HttpConnection`'s default credential discovery uses the 4.x-only `DefaultAWSCredentialsIdentityResolver` (building against 3.7.x fails with CS0103). The floor stays at `4.0.7.1`.

### Related Issues

Related to #968

### Check List

- [x] New functionality includes testing.


xluo-aws:
@Hailong-am , @SamuelCox could you help review this change?

Hailong-am:
Thanks for the fix — the SigV4 reimplementation itself looks correct:

- The manual `stringToSign` / signing-key derivation in `AwsSigV4Util.ComputeAuthorizationHeader` matches the spec exactly, and `ComputeAuthorizationHeaderMatchesPublishedAwsReferenceVector` pins it to AWS's own published reference vector — good, independent verification beyond "matches old output."
- The existing known-answer tests (`CanonicalRequestTests`, `AwsSigV4HttpConnectionTests`) still produce byte-identical signatures, so no regression for current callers.
- Removing all `Amazon.Runtime.Internal.*` usage in favor of `SHA256`/`HMACSHA256` + the public `AWSSDKUtils` helpers is good practice on its own merits — SigV4 is a small, stable, publicly documented algorithm, and .NET's own `System.Security.Cryptography` primitives carry a much stronger backward-compat guarantee than anything under an SDK's `Internal` namespace. This is a solid fix to ship regardless of the exact trigger conditions below.

One thing I want to mention transparently: I spent some time trying to reproduce the specific #968 symptom locally (matching `AWSSDK.Core 4.0.3.8`, `AWSSDK.OpenSearchServerless 4.0.5.5`, `net8.0`, real AOSS, and eventually the actual EKS Pod Identity credential-resolution path via `Amazon.Runtime.GenericContainerCredentials`), and wasn't able to get a 401 out of the pre-fix code in any of my setups — signatures came out correct across `AWSSDK.Core` 3.7.x and 4.x in every variant I tried. That's likely just a gap in my repro (EKS Pod Identity in a real cluster has details a local mock can't fully capture — token rotation timing, the actual agent's exact response shape, etc.), not a reason to doubt the report. If you have a minimal repro handy (even just the exact `AWSSDK.Core` patch + a stack trace/log snippet from the failure), I'd love to see it — mostly out of curiosity to close the loop on the mechanism, not as a blocker for this PR.

Separately, a packaging follow-up: the `.csproj` still floors `AWSSDK.Core` at `4.0.7.1` (unchanged by this PR). Since the new signing code only needs `SHA256`/`HMACSHA256`/public `AWSSDKUtils` — all available since 3.x — it'd be worth loosening that floor so 3.7.x consumers aren't forced to bump just to pick up this fix.

xluo-aws:
Thanks @Hailong-am — really appreciate the careful review, and especially that you tried to reproduce the mechanism.

I independently confirmed your finding. I set up the reporter's exact scenario (the released `OpenSearch.Net.Auth.AwsSigV4` 1.8.0 binary + `AWSSDK.Core 4.0.3.8` loaded at runtime, static and session-token credentials) and the pre-fix code produced **byte-identical, spec-correct signatures** with no `MissingMethodException`. So neither of us can get a 401 out of the signing path — because it was never a signing bug.

Chasing your packaging suggestion actually surfaced what I believe is the real mechanism. When I tried to lower the floor, the build fails against 3.7.x:

```
AwsSigV4HttpConnection.cs: error CS0103: The name
'DefaultAWSCredentialsIdentityResolver' does not exist in the current context
```

The 4.x bump (#931) changed the default credential-discovery line:

```diff
- _credentials = credentials ?? FallbackCredentialsFactory.GetCredentials();          // 3.7
+ _credentials = credentials ?? DefaultAWSCredentialsIdentityResolver.GetCredentials(); // 4.x
```

The reporter constructed the connection **without** explicit credentials, so 1.8.0 resolved via the legacy `FallbackCredentialsFactory`, which does **not** include the **EKS Pod Identity** provider (that lives in the newer resolver chain). That yields the wrong/missing identity → AOSS returns 401, while `awscurl`/boto3 and `aws sts get-caller-identity` succeed because they use modern resolution. That fits every symptom in the report, and it explains why the 401 isn't reproducible from the signer.

Two takeaways:

1. **Floor can't be loosened.** `AwsSigV4HttpConnection`'s default-credentials path hard-requires the 4.x-only `DefaultAWSCredentialsIdentityResolver`, so the `AWSSDK.Core` floor must stay at 4.x. I reverted the experiment. (The signing code alone would compile against 3.7, but the connection won't.)

2. **This PR is hardening, not the root fix for #968.** The actual Pod Identity fix is the credential-resolver switch already on `main` from #931. This PR's value is removing the fragile `Amazon.Runtime.Internal.*` dependency in favor of `System.Security.Cryptography` + public `AWSSDKUtils`, plus the added edge-case test coverage (published AWS reference vector, session token, non-UTC time, multi-value headers). I'll update the PR description to reflect that framing.

Still worth shipping on its own merits, but wanted to close the loop honestly on the mechanism.

Hailong-am (approved):

SamuelCox (approved):

PR #996, opened 2026-07-16
Migrate high-level serialization from Utf8Json to System.Text.Json (#388)

Hailong-am:
### Description

Adds a complete **System.Text.Json (STJ)** serialization path for `OpenSearch.Client` and `OpenSearch.Net`, alongside the existing vendored Utf8Json engine, as the substantive step of the migration tracked in #388.

**The default engine is unchanged** — Utf8Json remains the default, and STJ is opt-in via `.UseUtf8Json(false)` on `ConnectionSettings` or the `OSC_USE_STJ=true` environment variable. Existing behavior is fully preserved; users opt in when ready.

#### Highlights

- **STJ infrastructure** in `OpenSearch.Net` (serializer options, type-info resolver, low-level value converters) and `OpenSearch.Client` (`DefaultHighLevelSystemTextJsonSerializer`).
- **`InterfaceDataContractModifier`** — an *additive* `JsonTypeInfo` modifier that adjusts the default STJ contract to honor the existing `[DataMember]` / `[InterfaceDataContract]` data contracts without rebuilding the property set, so options-registered converters resolve normally at depth.
- **~90 domain converters** covering QueryDsl, Aggregations, Mapping, Analysis, Ingest, Search, Indices, Snapshot, Cluster, and Nodes.
- **`OpenSearch.Client`-native `InterfaceDataContractAttribute`** so high-level domain types declare their contract without taking a dependency on the vendored Utf8Json engine. Utf8Json's `MetaType` recognizes the marker by simple name, so the default engine still resolves contracts.
- **CI** runs the full unit suite under both engines and spot-checks integration on STJ (`test-jobs.yml`, `integration.yml`).

#### Testing

The STJ engine runs the **same** existing test suite via `OSC_USE_STJ`, so behavior parity is asserted by the full suite rather than by duplicating tests.

- Full unit suite green under **both** engines (System.Text.Json and Utf8Json).
- `Tests.Reproduce` green under both engines.

### Issues Resolved

Part of #388.

### Check List
- [x] New functionality includes testing (existing suite runs under both engines).
- [x] New functionality has been documented (CHANGELOG.md).
- [x] Commit changes are listed out in CHANGELOG.md.

### DIFF
--- a/src/OpenSearch.Client/Aggregations/AggregationContainer.cs
+++ b/src/OpenSearch.Client/Aggregations/AggregationContainer.cs
@@ -92,6 +92,17 @@ namespace OpenSearch.Client
 		}
 	}
 
+	internal class AggregationDictionaryFormatter : IJsonFormatter<AggregationDictionary>
+	{
+		private static readonly VerbatimDictionaryInterfaceKeysFormatter<string, IAggregationContainer> DictionaryKeysFormatter =
+			new VerbatimDictionaryInterfaceKeysFormatter<string, IAggregationContainer>();
+
+		public AggregationDictionary Deserialize(ref JsonReader reader, IJsonFormatterResolver formatterResolver) =>
+			new AggregationDictionary(DictionaryKeysFormatter.Deserialize(ref reader, formatterResolver));
+
+		public void Serialize(ref JsonWriter writer, AggregationDictionary value, IJsonFormatterResolver formatterResolver) =>
+			DictionaryKeysFormatter.Serialize(ref writer, value, formatterResolver);
+	}
 
 	[InterfaceDataContract]
 	[ReadAs(typeof(AggregationContainer))]
@@ -944,17 +955,4 @@ namespace OpenSearch.Client
 			return d;
 		}
 	}
-	internal class AggregationDictionaryFormatter : IJsonFormatter<AggregationDictionary>
-	{
-		private static readonly VerbatimDictionaryInterfaceKeysFormatter<string, IAggregationContainer> DictionaryKeysFormatter =
-			new VerbatimDictionaryInterfaceKeysFormatter<string, IAggregationContainer>();
-
-		public AggregationDictionary Deserialize(ref JsonReader reader, IJsonFormatterResolver formatterResolver) =>
-			new AggregationDictionary(DictionaryKeysFormatter.Deserialize(ref reader, formatterResolver));
-
-		public void Serialize(ref JsonWriter writer, AggregationDictionary value, IJsonFormatterResolver formatterResolver) =>
-			DictionaryKeysFormatter.Serialize(ref writer, value, formatterResolver);
-	}
-
-
 }
--- a/src/OpenSearch.Client/Aggregations/AggregationContainer.cs
+++ b/src/OpenSearch.Client/Aggregations/AggregationContainer.cs
@@ -84,25 +84,14 @@ namespace OpenSearch.Client
 
 		protected override string ValidateKey(string key)
 		{
-			if (AggregateFormatter.AllReservedAggregationNames.Contains(key))
+			if (AggregateConverter.AllReservedAggregationNames.Contains(key))
 				throw new ArgumentException(
-					string.Format(AggregateFormatter.UsingReservedAggNameFormat, key), nameof(key));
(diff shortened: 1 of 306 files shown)


src/OpenSearch.Client/CommonAbstractions/ConnectionSettings/ConnectionSettingsBase.cs
```diff
@@ -180,6 +185,27 @@ public TConnectionSettings DefaultFieldNameInferrer(Func<string, string> fieldNa
...
 
+		/// <summary>
+		/// Controls which serialization engine is used by the built-in high-level serializer.
+		/// When <c>true</c> (the default), the legacy vendored Utf8Json engine is used (honoring
+		/// <c>[JsonFormatter]</c> attributes). When <c>false</c>, the System.Text.Json engine is used
+		/// (honoring <c>[JsonConverter]</c> attributes).
+		/// <para></para>
+		/// The System.Text.Json engine can also be forced globally via the <c>OSC_USE_STJ</c> environment variable.
+		/// </summary>
+		public TConnectionSettings UseUtf8Json(bool use = true) => Assign(use, (a, v) => a._useUtf8Json = v);
```
xluo-aws: UseUtf8Json(false) fluent setter may not have effect, the serializer is built in the base constructor before any fluent call can run, and it keeps the engine choice permanently.

 new ConnectionSettings(pool).UseUtf8Json(false):

  1. The field initializer sets the flag from the env var only:
  private bool _useUtf8Json = ReadUtf8JsonDefault();
  2. The base constructor body then builds the serializer immediately:
  var defaultSerializer = new DefaultHighLevelSerializer(this);
  3. DefaultHighLevelSerializer(IConnectionSettingsValues) reads settings.UseUtf8Json once and freezes it into readonly fields that are never re-read (UsesUtf8Json => _formatterResolver != null).
  4. UseUtf8Json(bool) runs after construction and just mutates the settings flag:
  public TConnectionSettings UseUtf8Json(bool use = true) => Assign(use, (a, v) => a._useUtf8Json = v);

  By step 4 the serializer is already committed, and nothing rebuilds or re-reads the flag. So .UseUtf8Json(false) is a no-op for engine selection, and OSC_USE_STJ (read in step 1, before construction) is the only working switch.

src/OpenSearch.Client/Aggregations/AggregationContainer.cs
```diff
@@ -84,25 +84,14 @@ public static implicit operator AggregationDictionary(AggregationBase aggregator
 
 		protected override string ValidateKey(string key)
 		{
-			if (AggregateFormatter.AllReservedAggregationNames.Contains(key))
+			if (AggregateConverter.AllReservedAggregationNames.Contains(key))
```
Hailong-am, line 87: does this works for utf8json?

src/OpenSearch.Client/Aggregations/AggregationContainer.cs
```diff
@@ -955,4 +944,17 @@ Func<TAggregator, TAggregatorInterface> selector
...
+			new VerbatimDictionaryInterfaceKeysFormatter<string, IAggregationContainer>();
+
+		public AggregationDictionary Deserialize(ref JsonReader reader, IJsonFormatterResolver formatterResolver) =>
+			new AggregationDictionary(DictionaryKeysFormatter.Deserialize(ref reader, formatterResolver));
+
+		public void Serialize(ref JsonWriter writer, AggregationDictionary value, IJsonFormatterResolver formatterResolver) =>
+			DictionaryKeysFormatter.Serialize(ref writer, value, formatterResolver);
+	}
+
+
```
Hailong-am: format change?

src/OpenSearch.Client/CommonAbstractions/ConnectionSettings/ConnectionSettingsBase.cs
Hailong-am: Good call out, It make sense. UseUtf8Json will take effect. I will remove the setting.

Hailong-am:
close, in favor of #982

Hailong-am:
close it, we will use #1002

PR #1011, opened 2026-08-06
Add native OpenTelemetry tracing via System.Diagnostics.ActivitySource

### LINKED ISSUE #1010: [FEATURE] Native OpenTelemetry tracing via System.Diagnostics.ActivitySource
The .NET client currently instruments requests using the legacy System.Diagnostics.DiagnosticSource / DiagnosticListener mechanism (OpenSearch.Net.Diagnostics.DiagnosticSources). The OpenTelemetry SDK does not observe DiagnosticListener directly — it observes ActivitySource. As a result, users who want to trace their calls to OpenSearch cannot do so out of the box; they have to write custom glue code to bridge DiagnosticListener into OpenTelemetry, and the emitted data does not follow the standard OpenTelemetry semantic conventions (db.system, db.operation, etc.).

Add native OpenTelemetry tracing support to the client by introducing a System.Diagnostics.ActivitySource. On every request, the client would create a client-kind Activity (span) tagged with the standard OpenTelemetry database semantic conventions (https://opentelemetry.io/docs/specs/semconv/database/). Users would only need to subscribe the source in their tracer provider:

  Sdk.CreateTracerProviderBuilder()
      .AddSource("OpenSearch.Net.RequestPipeline")
      .AddOtlpExporter()
      .Build();

  No per-request code and no glue code required.

Bridging the existing DiagnosticListener events into OpenTelemetry on the consumer side. This pushes boilerplate onto every user, is error-prone, and still would not produce standard semantic-convention attributes. A native ActivitySource is the idiomatic .NET approach and matches how other .NET clients (e.g. the Elasticsearch .NET client) expose OpenTelemetry.

 - Because .NET has a first-class tracing primitive (ActivitySource) that the OpenTelemetry SDK observes natively, the .NET implementation is cleaner than in languages that require the OpenTelemetry SDK to create spans.
  - Reference: the Elasticsearch .NET client implements this in its transport layer.

YiyuanMiao:
### Description
  
  Requests to OpenSearch now emit an OpenTelemetry client span out of the box.
  Consumers just subscribe the source by name — no glue code:

  ```csharp
  Sdk.CreateTracerProviderBuilder()
      .AddSource("OpenSearch.Net.RequestPipeline")
      .Build();
 ```
Spans use the standard DB semantic conventions (db.system, db.operation, server.address, http.request.method, url.full, …) and are named after the REST operation (e.g. search). Opt-in and guarded by HasListeners, so no overhead when unused. The legacy DiagnosticSource mechanism is kept unchanged.

  Resolves #1010

 ### Changes
  
  Split into two commits — hand-written vs generated:
  
  - Commit 1 (hand-written): new ActivitySource + semantic-convention constants; instrumentation in RequestPipeline.CallOpenSearch(Async); OperationName added to IRequestParameters/RequestData; ApiGenerator template change; tests + docs.
  - Commit 2 (generated): _Generated/.../RequestParameters/* (36 files).

  ### ApiGenerator note (for reviewers)

  The span name / db.operation needs the REST operation name, which only lived as the [MapsApi] attribute. Instead of touching the hundreds of DoRequest call sites, the name rides on the existing RequestParameters → RequestData path. The template adds one line next to DefaultHttpMethod:

  cshtml
  public override string OperationName => "@names.RestSpecName";
  
  Regenerated with ./build.sh codegen --branch main --include-high-level --download → 481 overrides across 36 files. Base class defaults to null, so anything codegen doesn't cover (pings, sniffs, raw low-level calls) falls back to the HTTP method.

  Check List

  - [x] Tests added
  - [x] Documented + CHANGELOG updated
  - [x] Commits signed (DCO)


src/OpenSearch.Net/Diagnostics/OpenSearchClientActivitySource.cs
```diff
@@ -0,0 +1,42 @@
...
+	/// Holds the <see cref="ActivitySource"/> used to emit OpenTelemetry spans for requests to OpenSearch,
+	/// along with helpers to keep instrumentation cheap when nobody is listening.
+	/// </summary>
+	/// <remarks>
+	/// This is separate from the legacy <see cref="DiagnosticSources"/> / <see cref="DiagnosticListener"/>
+	/// mechanism, which is retained for backwards compatibility. OpenTelemetry SDKs only observe
+	/// <see cref="ActivitySource"/>, so a consumer only needs to add this source name to start collecting traces:
+	/// <code>AddSource(OpenSearchClientActivitySource.ActivitySourceName)</code>
+	/// </remarks>
+	public static class OpenSearchClientActivitySource
```
SamuelCox, line 26: Maybe I'm missing something, but the filename doesn't really match up with the class. `OpenTelemetry.cs` would make me think the class is for more than just the activity source.

src/OpenSearch.Net/Transport/Pipeline/RequestPipeline.cs
```diff
@@ -203,13 +206,15 @@ public async Task<TResponse> CallOpenSearchAsync<TResponse>(RequestData requestD
 			where TResponse : class, IOpenSearchResponse, new()
 		{
 			using (var audit = Audit(HealthyResponse, requestData.Node))
+			using (var activity = StartActivity(requestData))
```
SamuelCox, line 209: ANy reason we can't use the more modern `using var` now?

SamuelCox:
@YiyuanMiao A couple small comments but nothing major, looks good generally

YiyuanMiao:
@SamuelCox Thanks for the review! Renamed OpenTelemetry.cs → OpenSearchClientActivitySource.cs to match the class (b4f539a).
  
  On the using var suggestion: I kept the classic using (...) block to match the surrounding code — the two StartActivity lines I added just slot into existing stacked-using blocks, and it's the dominant style across the codebase (~89 classic using (...) blocks across 23 files vs. ~10 using var). Switching only this spot would make it inconsistent with the adjacent sync CallOpenSearch and the rest of RequestPipeline. Happy to modernize repo-wide, but that feels like a separate cleanup PR rather than part of the tracing change.
Do you prefer to change all the old form into modern ones within this PR? If so I'll immediately do this, otherwise I'll start a new pr, thx

SamuelCox:
> @SamuelCox Thanks for the review! Renamed OpenTelemetry.cs → OpenSearchClientActivitySource.cs to match the class ([b4f539a](https://github.com/opensearch-project/opensearch-net/commit/b4f539a3274d49b3371d039740c6ee2d8d98dea6)).
> 
> On the using var suggestion: I kept the classic using (...) block to match the surrounding code — the two StartActivity lines I added just slot into existing stacked-using blocks, and it's the dominant style across the codebase (~89 classic using (...) blocks across 23 files vs. ~10 using var). Switching only this spot would make it inconsistent with the adjacent sync CallOpenSearch and the rest of RequestPipeline. Happy to modernize repo-wide, but that feels like a separate cleanup PR rather than part of the tracing change. Do you prefer to change all the old form into modern ones within this PR? If so I'll immediately do this, otherwise I'll start a new pr, thx

No all good, makes sense. I tend to believe in incremental refactoring rather than having completely separate refactor PRs wherever possible (although obviously there are instances where the latter is necessary), but it's not too fussed about it ultimately. I'll approve

SamuelCox (approved):

PR #1017, opened 2026-08-10
Generate ML and SearchPipeline high-level client from OpenAPI spec

### LINKED ISSUE #1012: [FEATURE] Generate high-level client request/response types from the OpenAPI spec
## Is your feature request related to a problem?

The high-level client in `OpenSearch.Client` (typed request/response/descriptor + fluent and object-initializer syntax) is largely **hand-written**. New APIs and plugin surfaces — especially fast-moving ones like `ml`, `neural`, `security`, `flow_framework` — have to be modeled by hand to get first-class high-level support. That is slow, easy to get subtly wrong, and drifts from the published OpenAPI spec.

The `ApiGenerator` already generates the **low-level** client (`OpenSearch.Net`) and the high-level **base halves** (request URLs/paths, descriptors, client methods) from the spec, but it does **not** generate the high-level **request bodies and response types** (the typed POCOs, enums, and shared `$ref` types that make up the request/response payloads). Those are the parts still written by hand.

## What solution would you like?

Extend `ApiGenerator` to generate high-level **request bodies and response types** directly from the OpenAPI spec, so a namespace can be modeled end-to-end (URL + method + descriptor + typed request body + typed response + shared component types + enums) without hand-writing the payload shapes.

Design principle: **one schema, one owner (role-determined)** — every OpenAPI component schema is emitted by exactly one generator path, chosen by how it is used:

| Schema role | Emitted as |
|---|---|
| Shared object (nested `$ref` reused across ops) | POCO triple `X : IX` (+ descriptor); ops reference it as `IX` |
| Operation request body | request-body partial (`IXRequest` / `XRequest` / `XDescriptor`) |
| Operation 200 response | thin `XResponse : ResponseBase` embedding the data shape |
| Type already hand-written in OSC (e.g. `WriteResponseBase`) | mapped via registry to the existing C# type — not re-emitted |

Key generator capabilities this requires (all prototyped, see below):

- **Object `$ref` resolution.** Named object components must resolve to `I<Type>`, not degrade to `IDictionary<string,object>`. NSwag inlines `$ref`s (dropping `Reference.Id`) and NJsonSchema's `AllowAdditionalProperties` fires before ref resolution — solved with an object-schema reverse-map (the same trick already used for enums) plus reordering ref resolution ahead of the additional-properties fallback.
- **Collision-free dual paths.** The namespace path (shared POCOs) and the op path (`: ResponseBase` responses) must never both emit a type of the same name. A `PocoRegistry` with `skip` / `map` / `rename` rules enforces single ownership.
- **BCL / hand-written collision handling.** Generated type names that shadow BCL or existing OSC types (e.g. `Task`, `Action`, `Node`) are renamed via the registry.
- **Response envelope compatibility.** Generated response properties that collide with reserved `ResponseBase` wire names (`status`, `error`) are renamed in C# while keeping the wire name, matching the hand-written `ClusterHealthResponse` pattern, so server error envelopes still deserialize.

## What alternatives have you considered?

- **Continue hand-writing** high-level request/response types per API — the status quo; does not scale to the plugin surface and drifts from the spec.
- **Low-level only** for new APIs (callers use `OpenSearch.Net` with raw JSON) — loses the typed, discoverable high-level experience the client is valued for.

## Additional context

https://github.com/opensearch-project/opensearch-net/issues/668

A working proof of concept exists and has been validated end-to-end:

- **`ml.register_model_group`** generated request/response as the initial minimal (body-only, no-path-param) slice.
- **Full `ml` namespace** generated on top of it: shared POCO closure + request/response for all body operations, no-body ops via the existing generators, streaming ops excluded. Output compiles cleanly, is deterministic, and produces no cross-namespace churn (existing generated output byte-for-byte unchanged).
- **Integration-verified** against live ml-commons clusters across the CI matrix (OpenSearch 1.1.0 → 3.6.0), which surfaced and fixed real defects a unit-only approach would have missed (e.g. a `status`/`ResponseBase` envelope-deserialization bug), and drove per-API version guards where server endpoints did not yet exist.

Known limitations / follow-ups identified by the PoC:
- `oneOf` + discriminator (tagged union) shapes still need a custom `IJsonFormatter` (as ingest's `ProcessorFormatter` does); the attribute-only path does not cover them.
- Case-variant enum values that collapse to the same C# identifier are an inherent C#↔spec limitation and are documented as such.

This proposal is to land that generator capability upstream so it can be applied beyond `ml` to other namespaces/plugins.

Hailong-am:
## Description

Implements plugin-based model code generation for the `ml` and `search_pipeline` namespaces, producing typed request bodies, response classes, shared models (interface/class/descriptor triples), and enums from the OpenAPI specification.

### Key Features

**Generator Infrastructure (`src/ApiGenerator/`)**
- `IModelOverrides` interface + `ModelOverridesBase` for per-plugin configuration
- `ModelTypeResolver`: schema → C# type resolution with `GlobalMappedTypes` (Id, IndexName, Indices, IIndexSettings, etc.)
- `SchemaCatalog` recovery for NSwag-inlined `$ref` schemas (string-alias + oneOf unions)
- `PropertyTypeOverrides`: per-operation property-level type mapping for wrapper types (IAliases, IProperties, IDynamicTemplateContainer)
- `GlobalPropertyTypeOverrides`: schema-scoped overrides for shared types (TypeMapping.properties → IProperties)
- `MatchesNamespace`: supports both dotted namespaces and root-level operations
- `WrapperKeyUnionModel`: auto-detects oneOf wrapper-key discriminated unions
- `SuppressedUnionSchemaIds`: preserves hand-written code while classifying union schemas
- `SuppressLowLevelApiImport`: avoids naming collisions with low-level namespace imports

**Generated ML Output (`src/OpenSearch.Client/_Generated/ML/`)**
- ~72 ML operations with typed request/response/descriptor classes
- ~50 shared model types (interface + class + descriptor)
- ~20 string enums
- Rich C# types: Id, IndexName, Fields via GlobalMappedTypes

**Generated SearchPipeline Output (`src/OpenSearch.Client/_Generated/SearchPipeline/`)**
- 16 processor types (5 request, 9 response, 2 phase-results) via WrapperKeyUnion codegen
- PutSearchPipeline/GetSearchPipeline/DeleteSearchPipeline request/response types

**Tests**
- 92+ data-driven round-trip tests for ML request/response types
- 3 cross-language wire-compatibility tests proving JSON output matches opensearch-java
- Hand-written serialization tests (nested objects, enum wire values, oneOf)
- Integration tests with MinimumServerVersion auto-skip

**Samples**
- `create-model-aos`: Bedrock Titan connector + model registration on Amazon OpenSearch Service
- `Option<Uri>` with `CustomParser` for --host CLI (System.CommandLine has no built-in Uri support)
- `--aws` / `--aws-region` SigV4 signing options

### Verified Against Hand-Written Code

| Namespace | Fields Match | New Fields Caught | Known Limits |
|-----------|-------------|-------------------|--------------|
| snapshot.create | 5/5 ✅ | +1 (feature_states) | — |
| snapshot.restore | 8/8 ✅ | +4 (rename_alias_*, storage_type) | — |
| indices.create | 3/3 ✅ (mappings, settings, aliases) | — | 155 reachable types |
| indices.put_mapping | 10/10 ✅ (with PropertyTypeOverrides) | — | 2 spec gaps (runtime_fields, _size) |
| cluster.allocation_explain | 4/4 ✅ | +1 (current_node) | — |

### Issues Resolved

#1012

### Check List
- [x] New functionality includes testing
- [x] New functionality has been documented in the CHANGELOG
- [x] Commits are signed off


Hailong-am:
Preparing review...

Hailong-am:
**[PR Description](https://github.com/opensearch-project/opensearch-net/pull/1017)** updated to latest commit (https://github.com/opensearch-project/opensearch-net/commit/5087560a59bcf8490f7f9d89e9fc38ebfc9d9d01)

Hailong-am:
@SamuelCox @xluo-aws appreciate you could help to take a look at this PR

xluo-aws:
`ApiIntegrationTestBase.SkipForMinimumVersion` silently reports version-gated tests as **passed** instead of **skipped**, hiding the gating from CI output.

In `tests/Tests/Framework/EndpointTests/ApiIntegrationTestBase.cs`, the new `[I]` test overrides do:
```csharp
[I] public virtual async Task ReturnsExpectedStatusCode()
{
    if (SkipForMinimumVersion(out var reason)) { Console.WriteLine(reason); return; }
    await AssertOnAllResponses(...);
}
```
A bare `return` inside an `[I]`/`FactAttribute`-derived test is indistinguishable to xunit from "ran, nothing failed" — it reports as **Passed**, not **Skipped**. The `Console.WriteLine` reason is easy to miss in CI logs and won't show up in any pass/fail/skip summary.

This repo already has the right primitive for this: `[SkipVersion]`, consumed by `IntegrationTestDiscoverer` at discovery time so gated tests report as `Skipped` (visible, trackable, distinct from Passed). It's used in ~25+ existing test classes, and this PR's own `RegisterModelGroupApiTests` uses it correctly (`[SkipVersion("<2.8.0", "...")]`).

Since `MinimumServerVersion` is a compile-time-known literal baked into each generated `TInitializer` at codegen time, could the generator instead emit `[SkipVersion(...)]` directly on each generated integration test class (matching the hand-written convention), rather than reinventing a weaker runtime check in the shared base class? That would keep skip visibility consistent across the whole suite.

src/ApiGenerator/Domain/Code/HighLevel/Models/NamespaceModel.cs
```diff
@@ -0,0 +1,415 @@
...
+                    // Global schema-scoped property type override
+                    var overrideType = plugin.ResolvePropertyTypeOverride("", p.Key, id);
+                    if (overrideType != null) csharpType = overrideType;
+
+                    return new ModelProperty(
+                        WireName: p.Key,
+                        CsharpName: ToPascal(p.Key),
+                        CsharpType: csharpType,
+                        Type: typeRef,
+                        IsRequired: requiredNames.Contains(p.Key),
```
xluo-aws, line 122: is this attribute ever consumed?

xluo-aws:
The generated search_pipeline processor union types (IRequestProcessor, IResponseProcessor, IPhaseResultsProcessor) only carry the Utf8Json [JsonFormatter] attribute — no System.Text.Json converter exists anywhere in the repo for them, . Will consumer that use the STJ serializer backend who sets
RequestProcessors/ResponseProcessors on PutSearchPipeline get broken/wrong wire output?

SamuelCox:
> @SamuelCox @xluo-aws appreciate you could help to take a look at this PR

Yep will do shortly, apologies for not being around past week or so, have been on summer holidays.

SamuelCox:
I'll review this tomorrow or Monday as I don't have access to a laptop right now, and this pr is too large to review via mobile

SamuelCox:
By and large this looks good, it's a huge PR though and therefore difficult to review. In the future, maybe consider splitting the PRs or using this new feature https://github.github.com/gh-stack/ to make it more reviewable @Hailong-am

Hailong-am:
> By and large this looks good, it's a huge PR though and therefore difficult to review. In the future, maybe consider splitting the PRs or using this new feature https://github.github.com/gh-stack/ to make it more reviewable @Hailong-am 

~~Good suggestion, give a try in the future. For this pr there are lots of generated files that make it so large that could be a good candidate for stack.~~ Confirmed gh-stack don't support fork repo to use it at this moment https://github.com/github/gh-stack/issues/46 @SamuelCox

src/ApiGenerator/Domain/Code/HighLevel/Models/NamespaceModel.cs
Hailong-am: It's not used at this moment. It's safe to keep it, we may add validation for required fields in the future.

Hailong-am:
> `ApiIntegrationTestBase.SkipForMinimumVersion` silently reports version-gated tests as **passed** instead of **skipped**, hiding the gating from CI output.
> 
> In `tests/Tests/Framework/EndpointTests/ApiIntegrationTestBase.cs`, the new `[I]` test overrides do:
> 
> ```cs
> [I] public virtual async Task ReturnsExpectedStatusCode()
> {
>     if (SkipForMinimumVersion(out var reason)) { Console.WriteLine(reason); return; }
>     await AssertOnAllResponses(...);
> }
> ```
> 
> A bare `return` inside an `[I]`/`FactAttribute`-derived test is indistinguishable to xunit from "ran, nothing failed" — it reports as **Passed**, not **Skipped**. The `Console.WriteLine` reason is easy to miss in CI logs and won't show up in any pass/fail/skip summary.
> 
> This repo already has the right primitive for this: `[SkipVersion]`, consumed by `IntegrationTestDiscoverer` at discovery time so gated tests report as `Skipped` (visible, trackable, distinct from Passed). It's used in ~25+ existing test classes, and this PR's own `RegisterModelGroupApiTests` uses it correctly (`[SkipVersion("<2.8.0", "...")]`).
> 
> Since `MinimumServerVersion` is a compile-time-known literal baked into each generated `TInitializer` at codegen time, could the generator instead emit `[SkipVersion(...)]` directly on each generated integration test class (matching the hand-written convention), rather than reinventing a weaker runtime check in the shared base class? That would keep skip visibility consistent across the whole suite.

`SkipForMinimumVersion` has been removed and we can use existing `SkipVersion` attributes changes in commit https://github.com/opensearch-project/opensearch-net/pull/1017/changes/598e5aeee735b54177ea6e8a67a10e1499bff0f4

Hailong-am:
> The generated search_pipeline processor union types (IRequestProcessor, IResponseProcessor, IPhaseResultsProcessor) only carry the Utf8Json [JsonFormatter] attribute — no System.Text.Json converter exists anywhere in the repo for them, . Will consumer that use the STJ serializer backend who sets RequestProcessors/ResponseProcessors on PutSearchPipeline get broken/wrong wire output?

STJ converters added for those and commit in https://github.com/opensearch-project/opensearch-net/pull/1017/changes/598e5aeee735b54177ea6e8a67a10e1499bff0f4

xluo-aws:
Seems SearchPipeline STJ converters are generated but never wired up
WrapperKeyUnion.cshtml and PolicyWrapperKeyUnion.cshtml now emit a System.Text.Json.Serialization.JsonConverter<T> class per union (RequestProcessorConverter, ResponseProcessorConverter, PhaseResultsProcessorConverter), but none of them are registered in SystemTextJsonHighLevelSerializer's _options.Converters list as how SystemTextJsonHighLevelSerializer does for existing Converters.

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
...
 - Added native OpenTelemetry tracing via `System.Diagnostics.ActivitySource`, emitting a client span per request with standard semantic convention tags (`db.system`, `db.operation`, `server.address`, `server.port`, `http.request.method`, `http.response.status_code`, `url.full`); subscribe with `AddSource("OpenSearch.Net.RequestPipeline")` ([#1010](https://github.com/opensearch-project/opensearch-net/issues/1010))
+- Generate high-level `ml` and `search_pipeline` namespace clients (typed requests, responses, descriptors, shared models, and wrapper-key discriminated unions) from the OpenAPI spec via plugin-based model code generation (`--include-high-level`). Includes `GlobalMappedTypes` for rich C# types (Id, IndexName, Indices, IIndexSettings), `PropertyTypeOverrides` for wrapper types (IProperties, IAliases), cross-language wire-compatibility tests, and an AOS sample with SigV4 signing ([#1017](https://github.com/opensearch-project/opensearch-net/pull/1017))
```
xluo-aws: is the CrossLanguageWireCompatTests.cs reverted?

Hailong-am:
Good catch — you're right, the generated converters were dead code under the STJ backend. Fixed in `0c311c1`.

Rather than registering them in `SystemTextJsonHighLevelSerializer._options.Converters`, I made the templates (`WrapperKeyUnion.cshtml` / `PolicyWrapperKeyUnion.cshtml`) emit `[System.Text.Json.Serialization.JsonConverter(typeof(...))]` on the generated union interface, alongside the existing Utf8Json `[JsonFormatter]`:

```csharp
[JsonFormatter(typeof(RequestProcessorFormatter))]
[System.Text.Json.Serialization.JsonConverter(typeof(RequestProcessorConverter))]
public interface IRequestProcessor
```

Two reasons for the attribute over `_options.Converters`: (1) `RequestProcessors`/`ResponseProcessors` are `IList<I*Processor>`, and STJ skips `options.Converters` for collection elements — the attribute fires there where a registered converter wouldn't; (2) it auto-wires every current and future generated union, so this class of drift can't recur. Verified with the STJ unit-test suite (`OSC_USE_STJ=true`) passing.

CHANGELOG.md
Hailong-am: Yes, removed intentionally (commit `e431d1f`). The cross-language test asserted wire compatibility by comparing serialized output against fixtures from the other language clients, which was brittle and duplicated coverage we already have in-repo. The same guarantee — identical JSON field names and enum wire values — is covered by `MlGeneratedRoundTripTests` and `MlGeneratedSerializationTests`, which round-trip the generated types through both serializer backends. Happy to add it back if you'd prefer the explicit cross-language fixture check.

tests/Tests/ML/MlGeneratedRoundTripTests.cs
```diff
@@ -0,0 +1,240 @@
...
+			"UploadChunkRequest", "UploadChunkResponse",
+			"UploadModelRequest", "UploadModelResponse",
+			// Non-body op responses
+			"DeleteMemoryResponse",
+			"GetAgentResponse", "GetAgenticMemoryResponse",
+			"GetAllMemoriesResponse", "GetAllMessagesResponse", "GetAllToolsResponse",
+			"GetConnectorResponse", "GetControllerResponse",
+			"GetMemoryContainerResponse", "GetMemoryResponse",
+			"GetMessageResponse", "GetMessageTracesResponse",
+			"GetMlTaskResponse", "GetModelGroupResponse", "GetModelResponse",
```
xluo-aws: Should it be ML instead of Ml?

tests/Tests/ML/MLGeneratedRoundTripTests.cs
```diff
@@ -0,0 +1,240 @@
...
+			"GetAgentResponse", "GetAgenticMemoryResponse",
+			"GetAllMemoriesResponse", "GetAllMessagesResponse", "GetAllToolsResponse",
+			"GetConnectorResponse", "GetControllerResponse",
+			"GetMemoryContainerResponse", "GetMemoryResponse",
+			"GetMessageResponse", "GetMessageTracesResponse",
+			"GetMlTaskResponse", "GetModelGroupResponse", "GetModelResponse",
+			"GetProfileModelsResponse", "GetProfileResponse", "GetProfileTasksResponse",
+			"GetStatsResponse", "GetToolResponse",
+			"SearchModelGroupsResponse", "SearchModelsResponse",
+			"SearchResponse", "SearchTasksResponse",
```
xluo-aws, line 161: Seems missing GetProfileRequest, GetProfileModelsRequest, GetProfileTasksRequest, SearchModelsRequest, SearchTasksRequest

CHANGELOG.md
xluo-aws: No, you can just update the changelog to remove it.

tests/Tests/ML/MlGeneratedRoundTripTests.cs
Hailong-am: Good catch — the generated production types and the `_Generated/ML/` folder standardized on `ML` (`GetMLTask`, `MLTask`), so this test's `Ml*` naming was inconsistent. Renamed the class and helpers to `ML*` (`MLGeneratedRoundTripTests`, `MLRequestTypes`, `MLResponseTypes`, `FindMLTypes`, `IsMLGenerated`, `MLGeneratedNames`) and the files to `MLGenerated*Tests.cs`, and fixed a stale `GetMlTaskResponse` -> `GetMLTaskResponse` entry. Done in `623f699`.

tests/Tests/ML/MLGeneratedRoundTripTests.cs
Hailong-am: You're right, and there was a 6th missing too (`ProfileRequest`). Root cause: the test looked auto-discovering (reflection over the assembly) but then filtered through a hand-maintained `MLGeneratedNames` allowlist (`MLOperationPrefixes` was empty), so any newly generated request that wasn't hand-added silently fell out of coverage — a green suite proved nothing for it.

Fixed in `623f699` by adding all six (`GetProfileRequest`, `GetProfileModelsRequest`, `GetProfileTasksRequest`, `SearchModelsRequest`, `SearchTasksRequest`, `ProfileRequest`). All 46 generated `*Request` types are now covered — 99 ML tests pass under both the utf8json and STJ backends.

Note this allowlist is inherently drift-prone: the next codegen op will fall out again unless we add a CI invariant test ("every generated `*Request` is in the coverage set or explicitly excluded"). Happy to add that as a follow-up if you'd like.

CHANGELOG.md
Hailong-am: Done in `623f699` — removed the "cross-language wire-compatibility tests" phrase from the CHANGELOG entry, since those tests were dropped in `e431d1f`.

xluo-aws (approved):

SamuelCox (dismissed):

Hailong-am:
@SamuelCox I merged from main to resolve conflicts, could you approve it again?

SamuelCox (approved):

PR #935, opened 2025-06-18
Implement initial version of bulk stream APIs

Xtansia:
### Description
Implement initial version of bulk stream APIs

### Issues Resolved
Relates #916 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

Issue #916, opened 2025-03-25
[PROPOSAL] Release new version for .NET sdk (looking for the bulk streaming API))


dorki:
Hey team, it’s been 7 months since the last release (1.8.0, August 2024), and there are tons of commits piling up, including the Bulk Streaming API I really need. Can you please cut a new version so we can use these updates without building from source? Thanks!

BTW, in the Readme file there is a mention of a 'Secret' version 2.0.0 that i guess never made it out :)

Xtansia:
Hi @dorki,
Sorry for the delay in responding here, have been swamped with a number of other responsibilities.

The 2.0.0 release is certainly not "secret", it just hasn't been released yet, there is some breaking changes related to new code generation and the specification it's generated from has been in flux, so didn't want to cut it too early.

There is a large number of new APIs that have had "low-level" support generated, but don't necessarily have high-level wrappers & typings yet. The bulk streaming API is one such example, and given it only has a meaningful effect with the use of chunked transfer encoding, it would require some extra attention too ensure it's implemented correctly.

What would your expected user/dev experience of the bulk streaming API be? The same interface as the normal bulk API, in that it takes a list of operations and returns a materialized list of results, just doing chunked streaming behind the scenes, or are you envisaging some long-lived streaming enumerable/channel etc?

dorki:
I believe both approaches can be very beneficial. I would guess that the first approach would be easier to implement, and it would also be easier for users to test if it brings performance improvements over the regular bulk API,

I would be happy with a quick release with the first form (same interface as today), while for a long-term solution, I would prefer to have both.

dorki:
Hi @Xtansia, any news?

dorki:
hi there @Xtansia any news? its been 9 months already :(

Xtansia:
Hi @dorki, unfortunately, I'm no longer working on OpenSearch full-time, and as you can see from #935, we've been stalled due to a lack of other active maintainers to perform reviews.

xluo-aws:
@dorki , we are going to release 2.0 soon, the PR is merged recently and pending for release, which may take few weeks. https://github.com/opensearch-project/opensearch-net/issues/928. 
After that, we'll start looking at high level support of bulk streaming API that Xtansia already initiated and pick up the remaining work.

xluo-aws:
@dorki , 2.0 release is available now. https://www.nuget.org/profiles/opensearchproject

Hailong-am:
[...]

PR #1020, opened 2026-08-11
Enhance BulkAll with retries and affinity routing

yyfamazon:
### Description
Enhances the existing `BulkAll` helper (over `_bulk`) with opt-in, backward-compatible features, rather than shipping a separate helper.

Added to `BulkAll`:
- `DocumentAffinityKey(Func<T,string>)` — routes same-key docs to one worker, preserving per-key order; a slow/retrying key holds up only its own worker.
- Exponential-backoff-with-jitter retries via `RetryBaseDelay`/`RetryMaxDelay` (fixed `BackOffTime` stays the default; retry count still `BackOffRetries`).
- Fluent `WaitForActiveShards`, a `TotalDocumentsProcessed` counter, and `WorkerIndex` on the response.

Retained low-level `_bulk/stream` support (#935): request converter (`System.Text.Json` + Utf8Json) and newline-delimited response aggregation, usable directly via `client.BulkStream`. `BulkAll` itself always dispatches to `_bulk`.

An earlier opt-in `UseStreamingEndpoint()` on `BulkAll` was dropped after review: issuing one request/response per client-side batch doesn't leverage the streaming endpoint's server-side batching, so it was the same shape as `_bulk` with extra API surface.

Deferred (future work): true streaming (persistent connection + incremental writes) and an `IAsyncEnumerable<T>` source, once the low-level client supports them.

### Files
- `src/OpenSearch.Client/Document/Multiple/BulkAll/*` (observable, request/descriptor, observer, response, `RetryStrategy`)
- `src/OpenSearch.Client/Document/Multiple/BulkStream/*` (low-level `_bulk/stream` + STJ converter + response builder)
- Tests under `tests/Tests/Document/Multiple/BulkAll/` and `.../BulkStream/`

### Issues Resolved
Relates #916

### Check List
- [x] New functionality includes testing
- [x] Updated CHANGELOG


Hailong-am:
@yyfamazon would you mind to fix the failed CI

yyfamazon:
> @yyfamazon would you mind to fix the failed CI

Hi，this pr depends on https://github.com/opensearch-project/opensearch-net/pull/935. Since 935 hasn't be merged yet, so this pr fail. once 935 merged, ci will pass

Hailong-am:
> once 935 merged, ci will pass

@yyfamazon #935 merged, can you rebase/merge from main?

yyfamazon:
> > once 935 merged, ci will pass
> 
> @yyfamazon #935 merged, can you rebase/merge from main?

done. could you help click approve run ci test? thanks

src/OpenSearch.Client/Document/Multiple/BulkStreamAll/BulkStreamAllWorker.cs
```diff
@@ -0,0 +1,238 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using OpenSearch.Net;
+
+namespace OpenSearch.Client
+{
+	internal class BulkStreamAllWorker<T> where T : class
```
xluo-aws: Is this class consumed anywhere?

docs/design/bulk-stream-all.md
```diff
@@ -0,0 +1,460 @@
+# Design: BulkStreamAll — High-Level Streaming Bulk Ingestion Helper
```
xluo-aws, line 1: seems it describes a different, larger feature than what ships. It promises IAsyncEnumerable<T>,
  Channel<T>-based backpressure, MaxBatchBytes, FlushInterval, ChannelCapacity, FlushAsync()/IAsyncDisposable, and transparent _bulk fallback. None of these are implemented.

src/OpenSearch.Client/Document/Multiple/BulkStreamAll/BulkStreamAllObservable.cs
```diff
@@ -0,0 +1,395 @@
...
+						var wi = workerIndex;
+
+						// Apply backpressure
+						if (_request.BackPressure != null)
+							await _request.BackPressure.WaitAsync(_compositeCancelToken).ConfigureAwait(false);
+
+						pendingTasks.Add(BulkAsync(batch, page, 0, wi));
+
+						// Limit in-flight tasks
+						if (pendingTasks.Count >= _maxDegreeOfParallelism)
```
xluo-aws: Affinity ordering is unenforced: the in-flight throttle is a global count, allowing concurrent same-worker/same-key batches to reorder on the wire — the exact bug the feature claims to fix. AffinityPreservesOrderWithinSameKey does not catch this: sorting unique monotonic page numbers and asserting ascending is tautological. Fix by serializing dispatch to one in-flight batch per worker (per-worker await, not a global pendingTasks count), and rewrite the test to assert on server-side document _seq_no/version or observed completion order without pre-sorting.

src/OpenSearch.Client/Document/Multiple/BulkStreamAll/BulkStreamAllObservable.cs
```diff
@@ -0,0 +1,395 @@
...
+
+		private void BulkStreamAll(IObserver<BulkStreamAllResponse> observer)
+		{
+			var documents = _request.Documents;
+			var partitioned = new PartitionHelper<T>(documents, _bulkSize);
+
+			if (_affinityKeySelector != null)
+			{
+#pragma warning disable 4014
+				RunWithAffinityAsync(observer);
```
xluo-aws: BackPressure(...) is a silent no-op on the default (non-affinity) path: WaitAsync is only called in RunWithAffinityAsync (lines 128, 152), while the round-robin path
  (partitioned.ForEachAsync) never acquires. BulkAsync still calls Release() (280, 359), but ProducerConsumerBackPressure.Release() is guarded and no-ops when the semaphore is already
  full, so no exception and no throttling. Fix by acquiring on the round-robin path too (or pass the back-pressure through ForEachAsync). BackPressureThrottlesIngestion doesn't catch it —
  it only asserts completion + 0 failures, never that throttling occurred.

xluo-aws:
Affinity ordering is unenforced: the in-flight throttle (BulkStreamAllObservable.cs:133) is a global count, allowing concurrent same-worker/same-key batches to reorder on the wire —
  the exact bug the feature claims to fix. AffinityPreservesOrderWithinSameKey does not catch this: sorting unique monotonic page numbers and asserting ascending is tautological. Fix by
  serializing dispatch to one in-flight batch per worker (per-worker await, not a global pendingTasks count), and rewrite the test to assert on server-side document _seq_no/version or
  observed completion order without pre-sorting.

xluo-aws:
ContinueAfterDroppedDocuments defaults to false, contradicting the documented default and making the default path abort on a single bad doc. The POCO BulkStreamAllRequest.ContinueAfterDroppedDocuments is a plain bool (defaults false), and the observable does if (!_request.ContinueAfterDroppedDocuments) throw ThrowOnBadBulk(...).
  So out of the box, one non-retryable item (e.g., a mapping/parse rejection — a 400, not a 429) tears down the entire stream. The PR's own design doc states the default should be true("If true, continue processing after non-retryable failures (default true)").

src/OpenSearch.Client/Document/Multiple/BulkStreamAll/BulkStreamAllObservable.cs
yyfamazon: Fixed. Dispatch is now serialized per worker — each worker's next batch is chained after its previous batch completes (one in-flight batch per worker), instead of a global pendingTasks count, so same-key batches can't race on the wire. Rewrote the test to record each worker's completion order and assert strictly ascending page numbers without pre-sorting.

src/OpenSearch.Client/Document/Multiple/BulkStreamAll/BulkStreamAllObservable.cs
yyfamazon: Fixed. Both the affinity and round-robin paths now dispatch through one FlushAsync that awaits BackPressure.WaitAsync(...) before each batch, so it's applied on the default path too. The rewritten test measures peak concurrent bulk requests (via BufferToBulk for start / BulkResponseCallback for end) with slots set below MaxDegreeOfParallelism, so a no-op backpressure would now fail the assertion.

src/OpenSearch.Client/Document/Multiple/BulkStreamAll/BulkStreamAllWorker.cs
yyfamazon: Deleted the class

docs/design/bulk-stream-all.md
yyfamazon: Rewrote the doc to describe only the shipped feature and moved the unimplemented items (IAsyncEnumerable, MaxBatchBytes/FlushInterval, Channel<T> backpressure, FlushAsync()/IAsyncDisposable, transparent _bulk fallback) into an explicit "Future Work" section. Also removed the FlushInterval property since it was public API that did nothing.

yyfamazon:
> ContinueAfterDroppedDocuments defaults to false, contradicting the documented default and making the default path abort on a single bad doc. The POCO BulkStreamAllRequest.ContinueAfterDroppedDocuments is a plain bool (defaults false), and the observable does if (!_request.ContinueAfterDroppedDocuments) throw ThrowOnBadBulk(...). So out of the box, one non-retryable item (e.g., a mapping/parse rejection — a 400, not a 429) tears down the entire stream. The PR's own design doc states the default should be true("If true, continue processing after non-retryable failures (default true)").

Fixed. ContinueAfterDroppedDocuments now defaults to true, set in both BulkStreamAllRequest<T> (POCO constructor) and BulkStreamAllDescriptor<T> (descriptor constructor), so out of the box a non-retryable item (e.g. a 400 mapping/parse rejection) is routed to DroppedDocumentCallback and ingestion continues rather than tearing down the stream. Opt back into halt-on-first-drop with .ContinueAfterDroppedDocuments(false). Also corrected the interface XML-doc comment, which previously described the opposite (halt) behavior.

yyfamazon:
> Affinity ordering is unenforced: the in-flight throttle (BulkStreamAllObservable.cs:133) is a global count, allowing concurrent same-worker/same-key batches to reorder on the wire — the exact bug the feature claims to fix. AffinityPreservesOrderWithinSameKey does not catch this: sorting unique monotonic page numbers and asserting ascending is tautological. Fix by serializing dispatch to one in-flight batch per worker (per-worker await, not a global pendingTasks count), and rewrite the test to assert on server-side document _seq_no/version or observed completion order without pre-sorting.

Fixed. Dispatch is now serialized per worker rather than by a global pendingTasks count: each worker keeps a single "tail" task and chains its next batch after the previous one completes (await previousTail before scheduling), so at most one batch per worker is ever in flight and same-key batches can't race on the wire. Concurrency is still bounded by MaxDegreeOfParallelism (one in-flight batch × N workers).
 
 On the test: I went with the observed-completion-order option. Because each worker now completes its batches strictly in dispatch order and OnNext is serialized, the test records each worker's page numbers in arrival order and asserts 
they're strictly ascending without pre-sorting — so an out-of-order completion (the only way a same-worker reorder could surface) fails the assertion, rather than being masked by an OrderBy. Happy to also add a _seq_no/version-based 
 assertion for end-to-end server-applied ordering if you'd prefer that stronger check on top.

xluo-aws:
The per-worker chaining fixes the affinity ordering issue correctly, thanks. But it's applied to both producers — ProduceWithAffinityAsync (L131) and ProduceRoundRobinAsync (L133) share this FlushAsync — and on the round-robin path it costs throughput without buying anything.
The producer awaits flush, and flush awaits this specific worker's previous batch. With round-robin assignment (workerIndex = page % N, L155), page k+N always targets the same worker as page k, so the pipeline degenerates to: dispatch N batches → block on worker 0 → dispatch one → block on worker 1 → … Idle workers can't be given work while the producer is parked on one particular batch. The previous revision's Task.WhenAny let it proceed with whichever finished first, and that's also what BulkAllObservable does via ForEachAsync(Extensions.cs:241). It's most visible with retries: a batch backing off (up to RetryMaxDelay, default 30s) holds its worker, and the whole ingestion stalls behind it even though N-1 workers are free. Without a DocumentAffinityKey there's no ordering guarantee to protect, so the serialization is pure loss on the default path

SamuelCox:
I see there's a bunch of comments, can you ping me when it's ready to re-review please?

yyfamazon:
> The per-worker chaining fixes the affinity ordering issue correctly, thanks. But it's applied to both producers — ProduceWithAffinityAsync (L131) and ProduceRoundRobinAsync (L133) share this FlushAsync — and on the round-robin path it costs throughput without buying anything. The producer awaits flush, and flush awaits this specific worker's previous batch. With round-robin assignment (workerIndex = page % N, L155), page k+N always targets the same worker as page k, so the pipeline degenerates to: dispatch N batches → block on worker 0 → dispatch one → block on worker 1 → … Idle workers can't be given work while the producer is parked on one particular batch. The previous revision's Task.WhenAny let it proceed with whichever finished first, and that's also what BulkAllObservable does via ForEachAsync(Extensions.cs:241). It's most visible with retries: a batch backing off (up to RetryMaxDelay, default 30s) holds its worker, and the whole ingestion stalls behind it even though N-1 workers are free. Without a DocumentAffinityKey there's no ordering guarantee to protect, so the serialization is pure loss on the default path

you're right that per-worker serialization was pure loss on the round-robin path. Fixed: the round-robin producer now dispatches concurrently up to MaxDegreeOfParallelism and schedules by whichever batch finishes first (Task.WhenAny), the same way BulkAllObservable does via ForEachAsync. A slow or backing-off batch now only ties up one of the N in-flight slots instead of parking a whole worker, so the other N−1 keep ingesting. Per-worker chaining is now confined to RunWithAffinityAsync, where the ordering guarantee is what we're protecting. Backpressure is still applied on both paths

src/OpenSearch.Client/Document/Multiple/BulkStreamAll/BulkStreamAllObservable.cs
```diff
@@ -0,0 +1,458 @@
...
+				// Observe every scheduled dispatch so faulted batches never surface as unobserved task exceptions.
+				try { await Task.WhenAll(inFlight).ConfigureAwait(false); }
+				catch { /* the originating failure is surfaced by the caller */ }
+				throw;
+			}
+		}
+
+		// Affinity dispatch: hash each document's key to a worker and chain that worker's batches so at most one
+		// batch per worker is ever in flight, preserving on-the-wire ordering for operations sharing a key.
+		private async Task RunWithAffinityAsync(Func<IList<T>, long, int, Task> dispatch)
```
xluo-aws: RunWithAffinityAsync awaits FlushAsync inside the single-threaded foreach loop, and FlushAsync first awaits one specific worker's in-flight task:
  
  var previous = workerTails[workerIndex];
  if (previous != null)
      await previous.ConfigureAwait(false);   // blocks the *producer*, not just this worker's
  slot
  
  With a single enumerator, parking there halts intake for all workers, not just the lagging one.
  
  The backoff delay lives inside the very task being awaited: dispatch → BulkAsync →
  RetryDocuments → await Task.Delay(delay, ...). So one batch retrying stops the producer from
  reading another document for up to RetryMaxDelay × MaxRetries (30s × 3 by default) while the
  other N−1 workers sit idle. Same failure mode I flagged for round-robin ("a batch
  backing off holds its worker, and the whole ingestion stalls behind it even though N−1 workers
  are free") — the fix applied there didn't touch this path. Skewed affinity keys cause the same
  collapse without any failure: a hot key's worker flushes far more often, and each flush parks
  the producer while the others hold partial buffers.
  
  With uniform keys and no retries the stalls largely overlap and concurrency stays near N, so the
  happy path is fine. But retries under load are the scenario this PR is motivated by
  (Kafka/Kinesis), and there MaxDegreeOfParallelism > 1 silently degrades toward serial.
  Relatedly, the design doc's "Total in-flight batches are therefore bounded by
  MaxDegreeOfParallelism" is only an upper bound, not the typical value.
  
  Neither affinity test catches this: AffinityPreservesOrderWithinSameKey asserts ordering, which
  the current design satisfies precisely because it over-serializes, and
  BackPressureCapsInFlightRequestsWithAffinity pins in-flight to 1 slot. A test that injects a 429
  or latency on one key and asserts the other workers keep completing batches would pin it down.

src/OpenSearch.Client/Document/Multiple/BulkStreamAll/BulkStreamAllObservable.cs
```diff
@@ -0,0 +1,458 @@
...
+
+			if (!response.ApiCall.Success)
+				return await HandleBulkFailure(buffer, page, attempt, workerIndex, response).ConfigureAwait(false);
+
+			var retryableDocuments = new List<T>();
+			var droppedDocuments = new List<Tuple<BulkResponseItemBase, T>>();
+			var retryPredicate = _request.RetryDocumentPredicate ?? DefaultRetryPredicate;
+			var droppedCallback = _request.DroppedDocumentCallback ?? DefaultDroppedCallback;
+
+			foreach (var documentWithResponse in response.Items.Zip(buffer, Tuple.Create))
```
xluo-aws, line 416: BulkStreamAll only ever inspects the first response chunk — retry and dropped-document handling don't work for batches larger than one document. _bulk/stream returns newline-delimited JSON: one {took, errors, items} object per server-side batch, and batch_size defaults to 1 so Zip only examines document 0 only. Failures on documents 1..N−1 never reach RetryDocumentPredicate and never reach
  DroppedDocumentCallback, so ContinueAfterDroppedDocuments can't fire either. This affects both dispatch paths and undermines the PR's core value prop (retry, drop routing, per-batch reporting).

src/OpenSearch.Client/Document/Multiple/BulkStreamAll/BulkStreamAllObservable.cs
yyfamazon: Good catch — you're right that the earlier fix only covered the round-robin path and the affinity path still funneled everything through a single producer that awaited each worker's in-flight batch. Under retries that made
  MaxDegreeOfParallelism > 1 collapse toward serial, exactly as you describe.

  I've reworked RunWithAffinityAsync so the producer no longer awaits any dispatch. Each worker now has its own bounded queue drained by a dedicated consumer:

  - The producer only routes batches into queues[workerIndex] and blocks only when that one worker already holds AffinityWorkerQueueDepthDefault (2) unfinished batches — so a batch backing off holds up its own key alone; the other workers keep
    draining.
  - Each consumer dispatches strictly FIFO with at most one batch in flight, and the capacity slot is released only when a batch fully completes (retries included). So the ordering guarantee is unchanged — AffinityPreservesOrderWithinSameKey
    still holds — and a persistently slow key backpressures the source for that key rather than freezing intake.
  - Cancellation/failure semantics are preserved: first failure cancels the composite token, all consumers are observed via Task.WhenAll (no unobserved task exceptions), and queues are disposed in a finally. No new package dependency — it's
    built on two SemaphoreSlims.

  This brings the affinity path in line with the guarantee the round-robin path already provides. I'm also adding a regression test that blocks one affinity key's requests and asserts the other worker keeps completing batches (it
  deadlocks/times out against the old code and passes with the fix).

src/OpenSearch.Client/Document/Multiple/BulkStreamAll/BulkStreamAllObservable.cs
yyfamazon: Fixed — you're right that only the first chunk was ever parsed. _bulk/stream streams newline-delimited JSON (one {took, errors, items} object per server-side batch, batch_size defaulting to 1), but BulkStreamResponse was deserialized as a
  single object, so response.Items held only document 0 and every later document's failure skipped RetryDocumentPredicate / DroppedDocumentCallback / ContinueAfterDroppedDocuments.

  I fixed it with a custom BulkStreamResponseBuilder (CustomResponseBuilderBase, the same mechanism _msearch uses) wired in via RequestDefaults. It splits the streamed body into its individual objects, deserializes each with the configured
  serializer, and aggregates them — items concatenated in wire order, errors OR-ed, took summed — so retry, drop routing and per-batch reporting see every document regardless of batch_size/batch_interval or server-side chunking.

  I went with a response builder rather than a custom JsonFormatter deliberately: System.Text.Json's reader (8.0.6, no AllowMultipleValues) throws on a second top-level object, so it can't read NDJSON through JsonSerializer in one pass.
  Splitting into single objects keeps the fix correct under both the default Utf8Json engine and UseSystemTextJson(), with no serializer changes.

  Added tests that stream multiple chunks with a failure on a non-first document and assert both that all items are aggregated and that the dropped-document callback fires — verified they fail against the old single-object parsing and pass with the fix, on both engines.

xluo-aws:
Looks good now, my feedback has all been addressed. @SamuelCox , @Hailong-am please help review when you have time.

xluo-aws:
suggest to add following test cases:
   1. The retry/drop partition in BulkAsync has no live coverage. RetriesExhaustedThrowsException is [Skip]ped and AssertNonFirstFailureIsRouted covers a single non-retryable item on the ContinueAfterDroppedDocuments(true) path, so the three-way partition never really executes:
  
  - Partial 429 retry — item k returns 429, the rest 201. Assert the second request body contains only document k, the batch then succeeds, and Retries== 1 / TotalNumberOfRetries == 1. Nothing currently verifies the "only retryable items are re-sent" guarantee.
  - Retry exhaustion on persistent 429 — exactly MaxRetries + 1 requests, then an OpenSearchClientException matching "after retrying", with TotalNumberOfFailedBuffers == 1.
  - ContinueAfterDroppedDocuments(false) halts — assert it throws and that no later batch is dispatched.
  - Retry preserves affinity ordering — same key, batch 0 gets a 429 and backs off; batch 1 must still land after it. This is where the two headline features interact, and it's currently untested.
  
2. Aggregation builder edges. The generator always emits one item per chunk with LF endings, so some branches never run:
  
  - A chunk containing 2+ items (batch_size > 1), so items.AddRange runs with multiple elements.
  - Fewer response items than documents, to pin down whatever you decide Zip should do on a count mismatch. This is the cheapest guard against the original bug class returning.

yyfamazon:
> suggest to add following test cases:
> 
> 1. The retry/drop partition in BulkAsync has no live coverage. RetriesExhaustedThrowsException is [Skip]ped and AssertNonFirstFailureIsRouted covers a single non-retryable item on the ContinueAfterDroppedDocuments(true) path, so the three-way partition never really executes:
> 
> * Partial 429 retry — item k returns 429, the rest 201. Assert the second request body contains only document k, the batch then succeeds, and Retries== 1 / TotalNumberOfRetries == 1. Nothing currently verifies the "only retryable items are re-sent" guarantee.
> * Retry exhaustion on persistent 429 — exactly MaxRetries + 1 requests, then an OpenSearchClientException matching "after retrying", with TotalNumberOfFailedBuffers == 1.
> * ContinueAfterDroppedDocuments(false) halts — assert it throws and that no later batch is dispatched.
> * Retry preserves affinity ordering — same key, batch 0 gets a 429 and backs off; batch 1 must still land after it. This is where the two headline features interact, and it's currently untested.
> 
> 2. Aggregation builder edges. The generator always emits one item per chunk with LF endings, so some branches never run:
> 
> * A chunk containing 2+ items (batch_size > 1), so items.AddRange runs with multiple elements.
> * Fewer response items than documents, to pin down whatever you decide Zip should do on a count mismatch. This is the cheapest guard against the original bug class returning.

Added the coverage — thanks for the specific breakdown. All as deterministic [U] tests driven by an in-memory connection (no cluster), so the retry/drop partition finally has live coverage.

  Category 1 (BulkStreamAllRetryDropApiTests.cs):
  - Partial 429 retry — asserts the retry request body contains only the failing document, the batch then succeeds, and TotalNumberOfRetries == 1.
  - Retry exhaustion — exactly MaxRetries + 1 requests, then an OpenSearchClientException matching "after retrying", with TotalNumberOfFailedBuffers == 1.
  - ContinueAfterDroppedDocuments(false) halts — throws, and the second batch's documents never appear in any request.
  - Retry preserves affinity ordering — same key, batch 0 gets a 429 and backs off; asserts pages complete [0, 1].
  
  Category 2 (BulkStreamResponseParsingApiTests.cs, run under both the Utf8Json and System.Text.Json engines):
  - Chunk with 2+ items — exercises the multi-element items.AddRange.
  - Fewer response items than documents — pins the behavior: the response surfaces exactly what the server returned (3), never collapsing to the first chunk (1) nor padding to the document count (5).
  
  While in there I also added coverage the context implied was thin: custom RetryDocumentPredicate (both making a 400 retryable and suppressing a 429), the dropped-callback's item/document contents, ContinueAfterDroppedDocuments(true)
  continuation, BulkResponseCallback invocation count, TotalDocumentsProcessed, empty-stream, plus PartitionHelper batching and BulkStreamRequestFormatter NDJSON framing (both previously untested). Full unit suite is green.

docs/design/bulk-stream-all.md
```diff
@@ -0,0 +1,314 @@
...
+┌──────────────────────────────────────────────────────────────────┐
+│                    BulkStreamAllObservable<T>                      │
+│                                                                    │
+│  Producer loop                Per-worker dispatch (N = MaxDOP)     │
+│  ┌───────────────┐            ┌──────────┐ ┌──────────┐           │
+│  │ read source   │            │ worker 0 │ │ worker 1 │  ...       │
+│  │ route:        │──batch────▶│ 1 batch  │ │ 1 batch  │           │
+│  │  • hash(key)  │            │ in flight│ │ in flight│           │
+│  │  • or page%N  │            └────┬─────┘ └────┬─────┘           │
+│  └───────────────┘                 │            │                  │
```
Hailong-am: https://docs.opensearch.org/latest/api-reference/document-apis/bulk-streaming/
Why we need to have batch at client side? streaming provides server side batching size and interval, client side should not care about batching anymore with streaming. This sounds more like an enhancement to existing `BulkAll` API.

Hailong-am:
Thanks for the work here @yyfamazon — the client-side improvements (exponential backoff, document-ID affinity, per-worker queue isolation) are genuinely valuable. Two design-level observations:

### 1. This doesn't actually stream

The PR description says "built on the `_bulk/stream` endpoint" and the design doc describes streaming architecture, but the implementation is client-side batching — identical to `BulkAll`:

- `BulkStreamAllObservable` partitions documents into batches of `Size` (default 1000)
- Each batch is sent as a complete `BulkStreamAsync()` request-response call
- The HTTP connection is not kept open between batches
- Server-side `batch_size`/`batch_interval` parameters are not exposed or set

With the server's default `batch_size=1`, a 1000-doc batch produces 1000 separate NDJSON response objects that `BulkStreamResponseBuilder` reads entirely into memory, splits on newlines, and deserializes individually — more overhead than `_bulk`'s single JSON response, with no streaming benefit.

True streaming would require the low-level transport to keep the HTTP connection open and write docs incrementally (via `HttpCompletionOption.ResponseHeadersRead` + chunked writes), which `BulkStreamAsync` from #935 doesn't support.

### 2. Consider improving `BulkAll` instead of a new parallel type

The API surface is ~80% identical to `IBulkAllRequest`. The genuinely new features are:

- `MaxRetries` / `RetryBaseDelay` / `RetryMaxDelay` (exponential backoff)
- `DocumentAffinityKey` (affinity routing)
- `BulkResponseCallback`
- `WaitForActiveShards`

All of these are additive, backward-compatible additions to `IBulkAllRequest` — they don't require `_bulk/stream`. This avoids maintaining two ~550-line orchestrators (`BulkAllObservable` + `BulkStreamAllObservable`) with near-identical logic that will inevitably drift.

A possible path:
1. Add these improvements to `BulkAll` on `_bulk` (works on all clusters, stable API)
2. Add a `UseStreamingEndpoint()` option that switches the wire call to `_bulk/stream` when available
3. When the low-level client supports true streaming, add `IAsyncEnumerable<T>` source + incremental writes

docs/design/bulk-stream-all.md
yyfamazon: Agreed — you're right on both points. The _bulk/stream endpoint does its own server-side batching (batch_size/batch_interval), and a client-side batching layer that issues one request/response per batch isn't really "streaming"; it's the same shape as BulkAll.

  - Removed the separate BulkStreamAll helper entirely.
  - Folded its genuinely-new features into BulkAll as backward-compatible, opt-in options: DocumentAffinityKey (per-key ordered dispatch), exponential-backoff-with-jitter retries (RetryBaseDelay/RetryMaxDelay), a fluent WaitForActiveShards, and a TotalDocumentsProcessed counter.
  - Added UseStreamingEndpoint() on BulkAll — an opt-in that dispatches batches to _bulk/stream through a single orchestrator (no second ~550-line observable), keeping only the low-level _bulk/stream request/response support.

  So there's now one helper, and the streaming endpoint is an opt-in transport rather than a parallel API. I've updated this design doc to describe the BulkAll integration, and noted true streaming (persistent connection + incremental writes,
  once the low-level client supports it) and an IAsyncEnumerable<T> source as future work. Full unit suite passes on both serializer engines.

yyfamazon:
> Thanks for the work here @yyfamazon — the client-side improvements (exponential backoff, document-ID affinity, per-worker queue isolation) are genuinely valuable. Two design-level observations:
> 
> ### 1. This doesn't actually stream
> The PR description says "built on the `_bulk/stream` endpoint" and the design doc describes streaming architecture, but the implementation is client-side batching — identical to `BulkAll`:
> 
> * `BulkStreamAllObservable` partitions documents into batches of `Size` (default 1000)
> * Each batch is sent as a complete `BulkStreamAsync()` request-response call
> * The HTTP connection is not kept open between batches
> * Server-side `batch_size`/`batch_interval` parameters are not exposed or set
> 
> With the server's default `batch_size=1`, a 1000-doc batch produces 1000 separate NDJSON response objects that `BulkStreamResponseBuilder` reads entirely into memory, splits on newlines, and deserializes individually — more overhead than `_bulk`'s single JSON response, with no streaming benefit.
> 
> True streaming would require the low-level transport to keep the HTTP connection open and write docs incrementally (via `HttpCompletionOption.ResponseHeadersRead` + chunked writes), which `BulkStreamAsync` from #935 doesn't support.
> 
> ### 2. Consider improving `BulkAll` instead of a new parallel type
> The API surface is ~80% identical to `IBulkAllRequest`. The genuinely new features are:
> 
> * `MaxRetries` / `RetryBaseDelay` / `RetryMaxDelay` (exponential backoff)
> * `DocumentAffinityKey` (affinity routing)
> * `BulkResponseCallback`
> * `WaitForActiveShards`
> 
> All of these are additive, backward-compatible additions to `IBulkAllRequest` — they don't require `_bulk/stream`. This avoids maintaining two ~550-line orchestrators (`BulkAllObservable` + `BulkStreamAllObservable`) with near-identical logic that will inevitably drift.
> 
> A possible path:
> 
> 1. Add these improvements to `BulkAll` on `_bulk` (works on all clusters, stable API)
> 2. Add a `UseStreamingEndpoint()` option that switches the wire call to `_bulk/stream` when available
> 3. When the low-level client supports true streaming, add `IAsyncEnumerable<T>` source + incremental writes

Thanks — this was the right call, and I've reworked the PR to follow your suggested path.

  On "it doesn't actually stream": agreed. It was client-side batching with one discrete request/response per batch; batch_size/batch_interval weren't exposed, so at the server default (batch_size=1) a batch even came back as N NDJSON objects —more overhead than _bulk's single response, no streaming benefit. True streaming needs the low-level transport to hold the connection open and write incrementally, which BulkStreamAsync doesn't do yet.

  What I changed, against your three steps:

  1. Improvements added to BulkAll over _bulk — all opt-in and backward-compatible (no renames to the existing surface): DocumentAffinityKey (per-key ordered dispatch; a slow/retrying key holds up only its own worker),
     exponential-backoff-with-jitter retries via RetryBaseDelay/RetryMaxDelay (existing BackOffTime stays the default), a fluent WaitForActiveShards, and a TotalDocumentsProcessed counter. BulkResponseCallback already existed on BulkAll.
  2. UseStreamingEndpoint() — an opt-in that routes each batch to _bulk/stream through a single orchestrator (a SendBatchAsync seam), so there's one ~550-line observable, not two. The low-level _bulk/stream support is retained (including the STJ request converter and NDJSON response aggregation).
  3. IAsyncEnumerable<T> source + incremental writes — deferred to follow-up, gated on the low-level client supporting real streaming; noted in the design doc's Future Work.

  Net: the separate BulkStreamAll type is gone (nothing had shipped it), its features are additive BulkAll capabilities, and streaming is an opt-in transport. Tests re-homed onto BulkAll; full unit suite green on both the Utf8Json and
  System.Text.Json engines. Let me know if the API shape (names/defaults) looks right and I'll finalize.

src/OpenSearch.Client/Document/Multiple/BulkAll/BulkAllRequest.cs
```diff
@@ -189,6 +217,18 @@ public BulkAllRequest(IEnumerable<T> documents)
...
+		public Func<T, string> DocumentAffinityKey { get; set; }
+
+		/// <inheritdoc />
+		public TimeSpan? RetryBaseDelay { get; set; }
+
+		/// <inheritdoc />
+		public TimeSpan? RetryMaxDelay { get; set; }
+
+		/// <inheritdoc />
+		public bool UseStreamingEndpoint { get; set; }
```
Hailong-am: can we remove `UseStreamingEndpoint` at this moment? We're not leveraging the benefits of bulk streaming

yyfamazon: Agreed, and done — removed `UseStreamingEndpoint` (property, descriptor method, and its test); `BulkAll` now always dispatches to `_bulk`. You're right it wasn't leveraging streaming: one request/response per client-side batch is the same shape as `_bulk`, just extra API surface.

I kept the low-level `_bulk/stream` support from #935 (the request converter + newline-delimited response aggregation), since it's the generated endpoint binding plus serialization and is usable directly via `client.BulkStream`, independent of this helper. Updated the CHANGELOG and design doc to match, and noted a genuine streaming integration (persistent connection + incremental writes, leveraging server-side `batch_size`/`batch_interval`) as future work.

Hailong-am (dismissed):

src/OpenSearch.Client/Document/Multiple/BulkAll/BulkAllObservable.cs
```diff
@@ -90,23 +104,167 @@ public IDisposable Subscribe(BulkAllObserver observer)
...
+		// batch per worker is ever in flight and batches complete in the order they were enqueued.
+		private async Task ConsumeWorkerAsync(AffinityWorkerQueue queue, Func<IList<T>, long, int, Task> dispatch, int workerIndex)
+		{
+			while (true)
+			{
+				var next = await queue.DequeueAsync(_compositeCancelToken).ConfigureAwait(false);
+				if (next == null) break; // queue completed and drained
+
+				if (_partitionedBulkRequest.BackPressure != null)
+					await _partitionedBulkRequest.BackPressure.WaitAsync(_compositeCancelToken).ConfigureAwait(false);
```
xluo-aws: This WaitAsync can deadlock Reindex when DocumentAffinityKey is set.
  
ProducerConsumerBackPressure is directional: the producer acquires slots, the consumer repays them. The scroll side (ScrollAllObservable) waits once per page before pushing it downstream, and BulkAll has only ever called Release() — that's what the doc on IBulkAllRequest.BackPressure promises ("each bulk request will call Release"), and it's still what the round-robin path does. But here, the affinity worker acquires from that same semaphore before dispatching.
The math no longer balances. Producing one batch of Size docs costs the scroll producer Size/searchSize slots, and now the worker takes one more — but a completed bulk repays at most backPressureFactor. So whenever factor <= Size/searchSize, every batch drains the pool a little further. And that's a completely legal config — the funnelTooSmall guard in ReindexObservable happily accepts e.g. searchSize=10, Size=1000, maxConcurrency=4, factor=26. Eventually the pool hits zero with the scroll blocked in its WaitAsync, a worker holding a ready batch blocked right here, and nothing in flight. At that point nothing can ever call Release, and the reindex hangs forever.
The important part is that this used to be safe. Before this PR, a tight config just throttled: the consumer never needed a slot, so any in-flight bulk's Release() would wake the producer and things kept moving. Adding a second blocking acquirer on the same pool is what turns "slow" into "stuck". Even generous factors aren't fully safe — the producer is allowed to run the whole pool ahead in buffered pages, so with skewed keys there's a race where slots hit zero while a worker sits between dequeue and this line, with nothing in flight. Same terminal state, just timing-dependent.
And it's reachable without doing anything exotic: Reindex wires its shared back-pressure instance onto the user's request (bulkAllRequest.BackPressure = backPressure), and a DocumentAffinityKey set through ReindexRequest.BulkAll flows straight through.

yyfamazon: Good catch — confirmed and fixed in afea69bb6.

Root cause is exactly as you describe: `ProducerConsumerBackPressure` is directional. In a Reindex the scroll producer acquires a slot per page and the bulk side repays via `Release()` per completed batch — that's the contract the round-robin path has always honored (consumer only ever `Release`s). The affinity `ConsumeWorkerAsync` broke it by adding a *second* acquirer (`WaitAsync`) on the same pool while a batch still repays at most `backPressureFactor`. So for any config where a batch spans more pages than the factor repays — a legal setup that clears the `funnelTooSmall` guard, e.g. `searchSize=10, Size=1000, factor<100` — the pool drains monotonically until the scroll is parked in `WaitAsync`, this worker holds a ready batch it can't dispatch, and nothing is in flight to ever `Release`. Terminal.

Fix: removed the consumer-side acquire entirely. Per-worker concurrency is already bounded to one in-flight batch by `AffinityWorkerQueue`, so the affinity path now matches the round-robin path — it only `Release`s, never `WaitAsync`s. Ordering guarantees are unaffected.

Added a regression test (`AffinityDoesNotDeadlockWhenBackPressurePoolIsDrained`) that drains the back-pressure pool to zero — the state a buffered-ahead scroll produces — then runs an affinity `BulkAll` wired to that same pool the way `ReindexObservable` does. It hangs to the 30s timeout against the old code and completes in ~1s with the fix. Full unit suite green.

xluo-aws (approved):

SamuelCox (approved):

PR #1042, opened 2026-09-09
ci: bump peter-evans/create-pull-request to v7

Hailong-am:
## Problem / Motivation

The scheduled **Code Generation** workflow (`.github/workflows/code-gen.yml`) fails at the **Create Pull Request** step. Failed run: https://github.com/opensearch-project/opensearch-net/actions/runs/34070887129/job/101587882168

```
remote: Duplicate header: "Authorization"
The process '/usr/bin/git' failed with exit code 128
```

## Why it matters

This is the job that pulls the latest OpenSearch API spec and re-generates the client every Monday. While it is broken, the .NET client silently stops tracking the spec — no regen PR is opened, and the drift is invisible until someone notices by hand.

### DIFF
--- a/.github/workflows/code-gen.yml
+++ b/.github/workflows/code-gen.yml
@@ -77,7 +77,7 @@ jobs:
         run: echo "date=$(date +'%Y-%m-%d')" >> $GITHUB_OUTPUT
 
       - name: Create Pull Request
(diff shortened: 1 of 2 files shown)


Hailong-am:
@SamuelCox @joshuali925 could you help to review this tiny change

joshuali925 (approved):

PR #1047, opened 2026-09-14
Add discovered_cluster_manager to ClusterHealthResponse

### LINKED ISSUE #1044: [FEATURE] ClusterHealthResponse is missing discovered_master / discovered_cluster_manager fields
## Is your feature request related to a problem?

`ClusterHealthResponse` (high-level client, `OpenSearch.Client`) does not expose the `discovered_master` / `discovered_cluster_manager` fields that the raw `GET /_cluster/health` response includes. These fields indicate whether the cluster has successfully discovered a cluster manager (master) node, which is useful for diagnosing cluster formation / split-brain issues.

Because of this gap, consumers who need those fields are forced to fall back to the low-level client and parse the raw JSON response themselves, even though every other field of the cluster health response is otherwise available on the typed `ClusterHealthResponse`.

## What solution would you like?

Add `DiscoveredMaster` and `DiscoveredClusterManager` properties (mapped via `[DataMember(Name = "discovered_master")]` / `[DataMember(Name = "discovered_cluster_manager")]`, matching the existing pattern used for the other fields on `ClusterHealthResponse`) so the typed response fully represents the `GET /_cluster/health` payload.

## What alternatives have you considered?

- Falling back to the low-level client (`OpenSearch.Net`) and parsing the raw JSON body — works, but loses the typed, discoverable high-level API for what is otherwise a fully-typed response.
- Reproducing the shape manually in application code — brittle, and defeats the purpose of using the typed client.

## Additional context

Noticed while migrating an internal diagnostics tool from the low-level `StringResponse` API to the typed `ClusterHealthResponse` API; the missing fields forced us to keep using the raw response for that specific diagnostic path.

xluo-aws:
### Description

Adds `DiscoveredClusterManager` and `DiscoveredMaster` properties to the high-level client's `ClusterHealthResponse`, so the typed response fully represents the `GET /_cluster/health` payload. Previously, consumers needing these fields had to fall back to the low-level client and parse the raw JSON.

Both properties are nullable (`bool?`) because each field can legitimately be absent depending on the server version:

- `discovered_cluster_manager` was added in OpenSearch 2.0
- `discovered_master` is deprecated as of 2.0 and only returned by older servers or for backwards compatibility

Nullable keeps "not present" distinguishable from `false`, which matters for the diagnostics use case that motivated the issue. The property pair with doc remarks follows the existing precedent in `ClusterStateResponse` (`MasterNode` / `ClusterManagerNode`).

Also extends the existing `ClusterHealthApiTests` integration test with a version-gated assertion (`DiscoveredMaster` on pre-2.0 clusters, `DiscoveredClusterManager` on 2.0+), matching the pattern used in `ClusterStateApiTests`.

**Testing:**

- `OpenSearch.Client` and `Tests` projects build clean
- Verified deserialization through `InMemoryConnection` on both serializer engines (Utf8Json and System.Text.Json) across three payload shapes: both fields present (2.x/3.x), `discovered_master` only (1.x), and neither field — new properties populate when present and remain `null` when absent
- Integration test assertion will be exercised by the CI cluster matrix

### Issues Resolved

Closes #1044


Hailong-am (approved):
Looks good to me, eventually we should replace it with auto generated files.

PR #1050, opened 2026-09-16
Fix radial search (min_score/max_distance) dropping the k-NN and neural query clause

### LINKED ISSUE #1049: [BUG] Hybrid Query - Semantic KNN search - MinScore and MaxDistance failure
### What is the bug?
I am using the High Level OpenSearch.Net client to run a hybrid query.

In the semantic search section of the hybrid query, when I set up the KNN clause, I desire to use the MinScore() function:

```csharp
        //
	qrys => qrys.Knn(knn => 
                            knn.Field(fld => fld.ProductVector)
                                  .Vector(request.QueryVector?.ToArray() ?? [])
                                  .Filter(f => f.Bool(b => b.Filter(filter_clauses)))
                                  .MinScore(MIN_VECTOR_SCORE)) // high level client fails to create KNN
                                 // .K(20) //This works as expected
                                // .MaxDistance(1.0f) //tested MaxDistance as well and it also failed
```

When I use MinScore() or Max Distance() the high level client silently fails to include the KNN section in my query. 

I get a 400 response on the call and the Debug Information does not show the KNN section of the query in the #Request.

I end up getting an error in my normalization pipeline due to there only being one query present in the hybrid query.

Using K() works as expected.

Attempt to use MinScore or MaxDistance in a hybrid query as described here:
https://opensearch.org/blog/vector-radial-search/

I would expect the query to succeed.  I was able to run a low level query using MinScore in the OpenSearch dashboard with no and I got the desired results.  Attempting to replicate it in the high level client has failed.

Windows 11, .Net 10

I encountered the issue on version 2.0.0, so I tried bumping up to 2.1.0, but that did not resolve the issue.
OpenSearch.Client.JsonNetSerializer" Version="2.1.0"
OpenSearch.Net" Version="2.1.0"
OpenSearch.Net.Auth.AwsSigV4" Version="2.1.0"
OpenSearch.Client" Version="2.1.0"

Hailong-am:
### Problem

A `knn` or `neural` query that uses `min_score` or `max_distance` (radial search) instead of `k` was silently dropped from the request. The clause disappeared, so a hybrid query lost a sub-query and the server returned 400. Only `k` worked. `NeuralQuery` was also missing most of its documented fields.

### What changed

The "conditionless" check in both `KnnQuery` and `NeuralQuery` required `k`. A radial-search query has no `k`, so it looked empty and was removed before serialization. The check no longer requires `k` — a query is kept when it has a vector/input and a field.

`NeuralQuery` also gained the rest of its documented fields: `max_distance`, `min_score`, `query_image`, `filter`, `query_tokens`, `semantic_field_search_analyzer`, `method_parameters`, `rescore`, and `expand_nested_docs`.

### Tests

Added serialization unit tests for both queries covering `max_distance`-only and `min_score`-only bounds, and for `NeuralQuery` the new fields (including a `query_tokens`-only sparse query). Each fails on `main` and passes here.

Fixes #1049


SamuelCox (approved):

work-account637:
@SamuelCox 

Do you have any guidance on how long it will be before there is a release for this fix?

Hailong-am:
> @SamuelCox
> 
> Do you have any guidance on how long it will be before there is a release for this fix?

We plan next 2.2 release at Oct

Hailong-am:
@work-account637 2.2.0 is out, you can give it a try.

